PR #155, opened 2024-04-09
Add missing sinks spawn and spawnSync for shell injection

nadiraikido:
This PR adds the missing sinks `spawn` and `spawnAsync` for shell injection, and ensures that we only inspect the call if the `shell` option is `true` (by default, this is false in Node).

Additional information: https://nodejs.org/api/child_process.html#child_processspawncommand-args-options

### DIFF
--- a/library/sinks/ChildProcess.test.ts
+++ b/library/sinks/ChildProcess.test.ts
@@ -38,7 +38,7 @@ t.test("it works", async (t) => {
 
   agent.start([new ChildProcess()]);
 
-  const { exec, execSync } = require("child_process");
+  const { exec, execSync, spawn, spawnSync } = require("child_process");
 
   const runCommandsWithInvalidArgs = () => {
     throws(
@@ -48,6 +48,16 @@ t.test("it works", async (t) => {
 
     throws(
       () => execSync(),
+      /argument must be of type string. Received undefined/
+    );
+
+    throws(
+      () => spawn().unref(),
+      /argument must be of type string. Received undefined/
+    );
+
+    throws(
+      () => spawnSync(),
       /argument must be of type string. Received undefined/
     );
   };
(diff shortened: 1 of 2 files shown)


library/sinks/ChildProcess.ts
```diff
@@ -10,6 +10,17 @@ export class ChildProcess implements Wrapper {
     name: string,
     context: Context
   ): InterceptorResult {
+    // Ignore calls to spawn or spawnSync if shell option is not enabled
+    if (name === "spawn" || name === "spawnSync") {
+      const shellOption = args.find(
+        (arg) => typeof arg === "object" && arg !== null && "shell" in arg
```
hansott, line 16: maybe use the `isPlainObject` helper here?

library/sinks/ChildProcess.test.ts
```diff
@@ -80,4 +92,28 @@ t.test("it works", async (t) => {
...
+        ).unref(),
+      "Aikido runtime has blocked a Shell injection: child_process.spawn(...) originating from body.file.matches"
+    );
+
+    throws(
+      () =>
+        spawnSync(
+          "ls `echo .`",
+          [],
+          { shell: true },
```
hansott, line 113: Maybe set this to `"/bin/sh"` or something like that? To cover that path?

Including when it's set to `false`?

hansott (commented):
Good find!

library/sinks/ChildProcess.test.ts
nadiraikido: Should be resolved now, I have included several test cases for when `shell` is a string, false, etc.

library/sinks/ChildProcess.ts
```diff
@@ -10,6 +11,17 @@ export class ChildProcess implements Wrapper {
...
   ): InterceptorResult {
+    // Ignore calls to spawn or spawnSync if shell option is not enabled
+    if (name === "spawn" || name === "spawnSync") {
+      const shellOption = args.find(
+        (arg) => isPlainObject(arg) && "shell" in arg
+      );
+
+      if (!shellOption) {
+        return undefined;
+      }
```
hansott, line 22: If `shell` is `false`, then it will still continue to the checks below?

Screenshot 2024-04-09 at 17 45 13

`"shell" in arg` will return true if the property is available on the object but it doesn't check its value?

hansott (approved):

PR #219, opened 2024-05-31
Improve absolute path traversal protection

timokoessler:
Fixes bypasses like `/./etc/` or `/././etc/`

### DIFF
--- a/library/vulnerabilities/path-traversal/unsafePathStart.ts
+++ b/library/vulnerabilities/path-traversal/unsafePathStart.ts
@@ -22,12 +22,12 @@ const linuxRootFolders = [
 const dangerousPathStarts = [...linuxRootFolders, "c:/", "c:\\"];
 
 export function startsWithUnsafePath(filePath: string, userInput: string) {
-  const lowerCasePath = filePath.toLowerCase();
-  const lowerCaseUserInput = userInput.toLowerCase();
+  const normalizedPath = filePath.replace(/^(\/\.)*/g, "").toLowerCase();
+  const normalizedUserInput = userInput.replace(/^(\/\.)*/g, "").toLowerCase();
   for (const dangerousStart of dangerousPathStarts) {
     if (
-      lowerCasePath.startsWith(dangerousStart) &&
-      lowerCasePath.startsWith(lowerCaseUserInput)
+      normalizedPath.startsWith(dangerousStart) &&
+      normalizedPath.startsWith(normalizedUserInput)
     ) {
       return true;
     }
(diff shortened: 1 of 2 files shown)


library/vulnerabilities/path-traversal/unsafePathStart.ts
```diff
@@ -22,12 +24,24 @@ const linuxRootFolders = [
 const dangerousPathStarts = [...linuxRootFolders, "c:/", "c:\\"];
 
 export function startsWithUnsafePath(filePath: string, userInput: string) {
-  const lowerCasePath = filePath.toLowerCase();
-  const lowerCaseUserInput = userInput.toLowerCase();
+  // Check if path is relative (not absolute or drive letter path)
+  // Required because resolve will build absolute paths from relative paths
+  if (!/^(\/|\w:).*/.test(filePath)) {
+    return false;
```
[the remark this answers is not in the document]
timokoessler: Added

library/vulnerabilities/path-traversal/unsafePathStart.ts
```diff
@@ -22,12 +24,25 @@ const linuxRootFolders = [
...
+  if (!isAbsolute(filePath)) {
+    return false;
+  }
+
+  let origResolve = resolve;
+  // @ts-expect-error __wrapped is not typed
+  if (resolve.__wrapped) {
+    // @ts-expect-error Not type safe
+    origResolve = resolve.__original;
+  }
```
hansott, line 38: Maybe nice to add something like this in `wrap.ts`:

```ts
type WrappedFunction = Function & {
  __original: Function;
};

export function isWrapped(fn: unknown): fn is WrappedFunction {
  return (
    fn instanceof Function &&
    "__wrapped" in fn &&
    fn.__wrapped === true &&
    "__original" in fn &&
    fn.__original instanceof Function
  );
}

const { resolve } = require("path");
if (isWrapped(resolve)) {
  resolve.__original;
}
```

library/vulnerabilities/path-traversal/unsafePathStart.ts
```diff
@@ -22,12 +24,25 @@ const linuxRootFolders = [
...
+
+  const normalizedPath = origResolve(filePath).toLowerCase();
+  const normalizedUserInput = origResolve(userInput).toLowerCase();
```
hansott, line 41: is it possible that the resolve might throw if we pass weird user strings to it?

timokoessler: Also tested `resolve(...)` it with all kinds of strings, including emojis 😄, it did not throw.

hansott (approved):

PR #227, opened 2024-06-07
Add end2end tests for Next.js (standalone)

timokoessler:


library/sources/HTTPServer.test.ts
```diff
@@ -1,7 +1,7 @@
 import { wrap } from "../helpers/wrap";
-import * as pkg from "../helpers/isPackageInstalled";
```
hansott, line 2: Remove isPackageInstalled helper?

timokoessler: Done

sample-apps/nextjs-standalone/src/app/files/route.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  if (!stdout || stdout.length === 0) {
+    return NextResponse.json({ files: [] });
+  }
+
+  return NextResponse.json({
+    files: stdout.split("\n").filter((line) => line),
+  });
+}
```
hansott, line 36: maybe good to add a route handler with a database query?

.github/workflows/end-to-end-tests.yml
```diff
@@ -27,7 +27,7 @@ jobs:
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=5 -e MYSQL_ROOT_PASSWORD=mypassword -e MYSQL_DATABASE=catsdb --entrypoint sh mysql:8.0 -c "exec docker-entrypoint.sh mysqld --default-authentication-plugin=mysql_native_password"
         ports:
           - "27015:3306"
-    timeout-minutes: 5
+    timeout-minutes: 10
```
hansott, line 30: Looks like we can keep the timeout at 5?

sample-apps/nextjs-standalone/src/app/files/route.js
timokoessler: In addition to or instead of Shell Injection?

PR #233, opened 2024-06-12
Only discover endpoints/routes that exist

hansott:


library/sources/http-server/routeExists.test.ts
```diff
@@ -0,0 +1,12 @@
+import * as t from "tap";
+import { routeExists } from "./routeExists";
+
+t.test("it works", async () => {
+  t.same(routeExists(404), false);
+  t.same(routeExists(405), false);
+  t.same(routeExists(200), true);
+  t.same(routeExists(500), true);
```
hansott, line 8: @willem-delbare What if 500?

PR #241, opened 2024-06-14
Add support for GraphQL

timokoessler:
- Hooks into GraphQL `execute` and `executeSync`
- Adds a sample app
- Adds end2end tests


README.md
```diff
@@ -59,6 +59,7 @@ See list above for supported database drivers.
 * ✅ [`typeorm`](https://www.npmjs.com/package/typeorm)
 * ✅ [`bookshelf`](https://www.npmjs.com/package/bookshelf)
 * ✅ [`drizzle-orm`](https://www.npmjs.com/package/drizzle-orm)
+* ✅ [`graphql`](https://www.npmjs.com/package/graphql) 16.x
```
hansott, line 62: Let's add a new category: API tools? Query languages? Client-server protocols? API specs?

library/agent/Context.ts
```diff
@@ -18,6 +18,7 @@ export type Context = {
   user?: { id: string; name?: string };
   source: string;
   route: string | undefined;
+  graphql?: string[];
```
hansott, line 21: In `runWithContext`, I would overwrite this so that it gets reset?

library/sources/GraphQL.ts
```diff
@@ -0,0 +1,51 @@
...
+          userInputs.push(value);
+        }
+      }
+    }
+
+    if (userInputs.length > 0) {
+      if (Array.isArray(context.graphql)) {
+        context.graphql.push(...userInputs);
+      } else {
+        context.graphql = userInputs;
```
hansott, line 37: feels a bit weird to mutate the context here? of course, we normally wrap functions in a different way to call `runWithContext`

I think it's fine for now

PR #246, opened 2024-06-16
Add support for Fastify

timokoessler:
- [x] Add sample app
- [x] Wrap GET requests
- [x] Wrap other request methods and other functions
- [x] Patch default module function export - Adds a new wrapping method by hooking into require
- [x] Add tests
- [x] Add end2end tests

TODO: Investigate how require works with different versions (e.g. a package is installed multiple times)


library/agent/wrapRequire.ts
```diff
@@ -0,0 +1,61 @@
+import * as mod from "module";
+import { ModifyingRequireInterceptor } from "./hooks/ModifyingRequireInterceptor";
+import { defineProperty } from "../helpers/wrap";
+import { Agent } from "./Agent";
+const req = mod.prototype.require;
+
+const interceptors: ModifyingRequireInterceptor[] = [];
```
hansott, line 7: if we can rewrite this as a class, would be nice, so that we can reset state if a new instance is made? Similar to the other hooking

As we run this multiple times during tests

library/agent/wrapRequire.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const original = req.apply(this, [id]);
+
+    if (typeof original !== "function") {
+      throw new Error(
+        "Module must export a function to be wrapped during require"
+      );
+    }
+
+    if (original.__wrapped) {
```
hansott, line 25: Would prefer to keep this in state, which modules have been patched? Instead of setting __wrapped?

library/agent/hooks/WrappableRequireSubject.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    const version = getPackageVersion(versionedPackage.getName());
+    if (!version) {
+      return;
+    }
+
+    if (!satisfiesVersion(versionedPackage.getRange(), version)) {
+      return;
+    }
```
hansott, line 28: prefer that this is checked in applyHooks like the others?

timokoessler:
See https://github.com/AikidoSec/firewall-node/pull/359

PR #262, opened 2024-07-02
Refactor file system function hooking

hansott:
We forgot `link` in the first list

This structure makes it less likely to make mistakes.

### DIFF
--- a/library/sinks/FileSystem.ts
+++ b/library/sinks/FileSystem.ts
@@ -43,6 +43,9 @@ const functions: Record<string, FileSystemFunction> = {
   copyFile: { pathsArgs: 2, sync: true, promise: true },
   cp: { pathsArgs: 2, sync: true, promise: true },
   link: { pathsArgs: 2, sync: true, promise: true },
+  watch: { pathsArgs: 1, sync: false, promise: false },
+  watchFile: { pathsArgs: 1, sync: false, promise: false },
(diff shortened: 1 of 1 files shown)


library/sinks/FileSystem.ts
```diff
@@ -4,50 +4,46 @@ import { InterceptorResult } from "../agent/hooks/MethodInterceptor";
...
+  symlink: { pathsArgs: 2, sync: true, promise: true },
+  stat: { pathsArgs: 1, sync: true, promise: true },
+  statfs: { pathsArgs: 1, sync: true, promise: true },
+  truncate: { pathsArgs: 1, sync: true, promise: true },
+  utimes: { pathsArgs: 1, sync: true, promise: true },
+  writeFile: { pathsArgs: 1, sync: true, promise: true },
+  copyFile: { pathsArgs: 2, sync: true, promise: true },
+  cp: { pathsArgs: 2, sync: true, promise: true },
+  link: { pathsArgs: 2, sync: true, promise: true },
+};
```
timokoessler, line 46: Maybe add [glob](https://nodejs.org/api/fs.html#fspromisesglobpattern-options), [watch](https://nodejs.org/api/fs.html#fspromiseswatchfilename-options) and [mkdtemp](https://nodejs.org/api/fs.html#fspromisesmkdtempprefix-options)?

PR #275, opened 2024-07-11
Check for min Node.js version

timokoessler:


library/helpers/getNodeVersion.ts
```diff
@@ -0,0 +1,7 @@
+export function getNodeVersionStr(): string {
+  return process.version;
```
hansott, line 2: maybe this function is a bit too much? also not used outside? ^^

library/agent/Agent.ts
```diff
@@ -375,6 +376,10 @@ export class Agent {
       throw new Error("Agent already started!");
     }
 
+    if (getMajorNodeVersion() < 16) {
+      throw new Error("Node.js version 16 or higher is required");
+    }
```
hansott, line 381: Maybe just print something to the console instead of just crashing the app?

PR #281, opened 2024-07-15
Add support for micro server

hansott:


library/sources/HTTPServer.ts
```diff
@@ -47,28 +46,21 @@ export class HTTPServer implements Wrapper {
...
-      .modifyArguments("on", (args, subject, agent) => {
-        return this.wrapOn(args, "https", agent);
-      });
+    ["http", "https"].forEach((module) => {
+      hooks
+        .addBuiltinModule(module)
+        .addSubject((exports) => exports)
+        .modifyArguments("Server", (args, subject, agent) => {
+          return this.wrapRequestListener(args, module, agent);
+        })
```
hansott, line 55: This is new, even though Server looks like a class, it's a function and we need to wrap the listener that is passed here: https://github.com/vercel/micro/blob/main/packages/micro/src/bin/micro.ts#L146

PR #293, opened 2024-07-18
Decode buffers passed as path to fs functions

timokoessler:


timokoessler:
This pr also changes the Node.js version in unit tests to `22.4.x` because of the following nom bug: https://github.com/npm/cli/issues/7657

PR #307, opened 2024-07-29
Add more context in metadata

hansott:


library/vulnerabilities/shell-injection/checkContextForShellInjection.ts
```diff
@@ -30,7 +30,9 @@ export function checkContextForShellInjection({
           kind: "shell_injection",
           source: source,
           pathToPayload: path,
-          metadata: {},
+          metadata: {
+            command: command,
```
hansott, line 34: Might contain sensitive information though? (if it's a false positive)

library/vulnerabilities/sql-injection/checkContextForSqlInjection.ts
```diff
@@ -33,7 +33,9 @@ export function checkContextForSqlInjection({
           kind: "sql_injection",
           source: source,
           pathToPayload: path,
-          metadata: {},
+          metadata: {
+            sql: sql,
```
hansott, line 37: Might contain sensitive information though? (if it's a false positive)

PR #312, opened 2024-08-01
Log if no token and blocking is disabled

timokoessler:


library/agent/Agent.ts
```diff
@@ -387,6 +388,12 @@ export class Agent {
       this.logger.log("Found token, reporting enabled!");
     } else {
       this.logger.log("No token provided, disabling reporting.");
+
+      if (!this.block && !isAikidoCI()) {
+        console.log(
+          "Aikido: Running in monitoring only mode without reporting to Aikido Cloud"
```
hansott, line 394: Maybe also mention: Use `AIKIDO_BLOCK=true` to enable blocking mode.

library/agent/Agent.ts
```diff
@@ -387,6 +388,13 @@ export class Agent {
       this.logger.log("Found token, reporting enabled!");
     } else {
       this.logger.log("No token provided, disabling reporting.");
+
+      if (!this.block && !isAikidoCI()) {
+        // eslint-disable-next-line no-console
+        console.log(
+          "Aikido: Running in monitoring only mode without reporting to Aikido Cloud. Set AIKIDO_BLOCK=true to enable blocking."
```
hansott, line 395: Keep similar to `console.log(`AIKIDO: ${message}`);`?

PR #313, opened 2024-08-02
Convert arrays to plain strings in Context

timokoessler:
This fixes
- Bypass using HTTP Parameter pollution
- Bypass by unexpected arrays in JSON without sufficient validation of user input

```js
db.exec(`INSERT INTO cats_2 (petname, comment) VALUES ('${req.query.petname}');`)
```
If req.query.petname is an array, it will be automatically converted into a comma separated string. This can for example be used to insert multiple rows into the db by `?petname='&petname=1)&petname=(2`.


timokoessler:
Benchmark shows that `.join()` is 28% faster than using template literals and 4-5% faster than using `.toString()`.

PR #323, opened 2024-08-07
Use endpoint with lowest threshold

hansott:


library/ratelimiting/getRateLimitedEndpoint.ts
```diff
@@ -0,0 +1,26 @@
...
+  const matches = config
+    .getEndpoints(context)
+    .filter((m) => m.rateLimiting && m.rateLimiting.enabled);
+
+  if (matches.length === 0) {
+    return undefined;
+  }
+
+  matches.sort((a, b) => {
+    if (a.rateLimiting.windowSizeInMS === b.rateLimiting.windowSizeInMS) {
```
timokoessler, line 18: If we have two matching routes, e.g.:
- /test/* - max. 5 requests in 30s
- /test/v2/* - max. 2 requests in 40s

In this case we will prefer the request with the smaller windows size? Maybe we should use a metric that includes the window size and max requests?

timokoessler (approved):

PR #324, opened 2024-08-08
Implement new hook system

timokoessler:
This PR introduces a new hooking system. Instead of checking if a package is installed and patching it directly, we now wrap the require function and patch modules on their first import.

## Advantages

- Only patch modules that are used
- Allows modifying default exports (required for Fastify, Undici, Koa)
- Should improve support for bundlers
- Do not wrap a method multiple times if different interceptor types are used
- Support patching multiple versions of one package, if it's installed directly and indirectly with a different version as a sub dependency

## Tasks

- [x] Wrap require calls
- [x] Support patching internal modules and main exports of packages
- [x] Support js file export wrapping
- [x] Add support for wrapping new instances
- [x] Wrap globals using new hook system
- [x] Wrap process.getBuiltinModule
- [x] Migrate all sources and sinks to new api


library/agent/Agent.ts
```diff
@@ -438,10 +430,30 @@ export class Agent {
       });
   }
 
-  onFailedToWrapMethod(module: string, name: string) {
+  onFailedToWrapMethod(module: string, name: string | undefined) {
     this.logger.log(`Failed to wrap method ${name} in module ${module}`);
   }
 
+  onFailedToWrapModule(module: string, error: Error) {
```
hansott, line 437: Why does `onFailedToWrapMethod` accept `string | undefined` if we have this one?

library/agent/hooks/VersionedPackage.ts
```diff
@@ -15,25 +14,57 @@ export class VersionedPackage {
...
-  addSubject(selector: (exports: any) => unknown): WrappableSubject {
-    const fn = new WrappableSubject(selector);
-    this.subjects.push(fn);
+  onFileRequire(relativePath: string, interceptor: RequireInterceptor) {
+    if (typeof relativePath !== "string") {
+      throw new Error("Relative path must be a string");
+    }
+    if (typeof interceptor !== "function") {
+      throw new Error("Interceptor must be a function");
+    }
```
hansott, line 32: Doesn't TypeScript enforce this? Same for relative path? or does this come from externally?

library/agent/hooks/VersionedPackage.ts
```diff
@@ -15,25 +14,57 @@ export class VersionedPackage {
...
+      );
+    }
+
+    if (relativePath.startsWith("./")) {
+      relativePath = relativePath.slice(2);
+    }
 
-    return fn;
+    this.requireFileInterceptors.set(relativePath, interceptor);
+    // Allow chaining
```
hansott, line 59:
```suggestion

```

library/agent/hooks/isBuiltinModule.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function isBuiltinModule(moduleName: string) {
+  // Added in Node.js v18.6.0, v16.17.0
+  if (typeof mod.isBuiltin === "function") {
+    return mod.isBuiltin(moduleName);
+  }
+
+  // The modulelist does not include the node: prefix
+  if (moduleName.startsWith("node:")) {
+    return moduleList.includes(moduleName.slice(5));
```
hansott, line 17: removeNodePrefix?

library/agent/Agent.ts
timokoessler: Because if a default export (root, not esm .default) is wrapped, there is no methodName. Can change to print default export instead.

library/agent/hooks/isBuiltinModule.test.ts
```diff
@@ -0,0 +1,12 @@
...
+import { isBuiltinModule } from "./isBuiltinModule";
+
+t.test("it works", async (t) => {
+  t.equal(isBuiltinModule("fs"), true);
+  t.equal(isBuiltinModule("mysql"), false);
+  t.equal(isBuiltinModule("http"), true);
+  t.equal(isBuiltinModule("node:http"), true);
+  t.equal(isBuiltinModule("test"), false);
+  // @ts-expect-error Testing the undefined case
+  t.equal(isBuiltinModule(undefined), false);
```
hansott, line 11:
```suggestion
  t.equal(isBuiltinModule(""), false);
```

Maybe better to check empty string, we would never be able to pass undefined to this function (typescript)?

PR #331, opened 2024-08-19
Change SSRF redirect test urls

timokoessler:


library/sinks/Fetch.test.ts
```diff
@@ -140,12 +150,10 @@ t.test(
       {
         ...context,
         // Redirects to http://127.0.0.1/test
-        ...{ body: { image: "https://dub.sh/aikido-ssrf-test" } },
+        ...{ body: { image: redirectUrl.ip } },
```
hansott, line 153: should we move the comments to the def at the top?

timokoessler: Moved them.

hansott (approved):

PR #343, opened 2024-08-27
Wrap IPv6 with square brackets so that we can construct URL

hansott:
given an URL like

http://[::]:80

axios and follow redirects will pass `::` as hostname to http(s).request

we try to construct an URL object from these separate arguments (protocol, hostname, port, path, ...) but this fails because we end up with the following URL:

http://:::80

so we need to wrap it again with square brackets if it's an IPv6 address


library/sinks/http-request/getUrlFromHTTPRequestArgs.ts
timokoessler, line 1: Should we maybe call `wrapWithSquareBracketsIfNeeded` in the function `mergeURLWithRequestOptions` too?

timokoessler (approved):

PR #349, opened 2024-08-29
SQL tokenizer

hansott:


Makefile
```diff
@@ -1,3 +1,8 @@
+SQL_TOKENIZER = sql-tokenizer/pkg/sql_tokenizer.js
+
+sql-tokenizer/pkg/sql_tokenizer.js:
```
foca, line 3: Should this depend on `sql-tokenizer/**/*.rs` to avoid building if the rust code hasn't changed?

hansott: Fixed!

hansott:
See https://github.com/AikidoSec/zen-internals

PR #356, opened 2024-09-04
Enable firewall only if env vars are set

timokoessler:
If none of the following environment variables are defined, the firewall returns as soon as possible and no longer changes any imported packets.

- AIKIDO_BLOCKING
- AIKIDO_BLOCK
- AIKIDO_TOKEN
- AIKIDO_DEBUG

### DIFF
--- a/library/helpers/shouldEnableFirewall.ts
+++ b/library/helpers/shouldEnableFirewall.ts
@@ -1,0 +1,29 @@
+import { isDebugging } from "./isDebugging";
+import { shouldBlock } from "./shouldBlock";
+
+/**
+ * Only enable firewall if at least one of the following environment variables is set to a valid value:
+ * - AIKIDO_BLOCKING
+ * - AIKIDO_BLOCK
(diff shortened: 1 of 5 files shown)


hansott (approved):

library/helpers/shouldEnableFirewall.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  if (isDebugging()) {
+    return true;
+  }
+
+  if (!isAikidoCI()) {
+    // eslint-disable-next-line no-console
+    console.log(
+      "AIKIDO: Firewall is disabled. Configure one of the following environment variables to enable it: AIKIDO_BLOCK, AIKIDO_TOKEN, AIKIDO_DEBUG."
```
hansott, line 33:
```suggestion
      "AIKIDO: Zen is disabled. Configure one of the following environment variables to enable it: AIKIDO_BLOCK, AIKIDO_TOKEN, AIKIDO_DEBUG."
```

PR #360, opened 2024-09-09
Support http2 client (SSRF)

timokoessler:
- [x] Detect direct SSRF attacks
- [x] Check dns lookup to detect SSRF with domain
- [x] DNS lookup exception is somehow catched in Node.js v16
- [ ] Check for SSRF redirect attacks


timokoessler:
See #372

PR #362, opened 2024-09-11
Add support for koa

timokoessler:
Adds support for https://www.npmjs.com/package/koa
Used by Strapi, for example.
Requires new hook system #324


library/middleware/koa.ts
```diff
@@ -0,0 +1,37 @@
+import { shouldBlockRequest } from "./shouldBlockRequest";
+import { escapeHTML } from "../helpers/escapeHTML";
+import type Application = require("koa");
```
hansott, line 3: not possible with

`import type * as Application from "koa"`? :D

library/sinks/Path.test.ts
```diff
@@ -109,9 +109,7 @@ t.test("it works", async (t) => {
     const error = t.throws(() => join("/etc/", "test.txt"));
     t.same(
       error instanceof Error ? error.message : null,
-      getMajorNodeVersion() <= 22
-        ? "Zen has blocked a path traversal attack: path.normalize(...) originating from body.file.matches"
-        : "Zen has blocked a path traversal attack: path.join(...) originating from body.file.matches"
+      "Zen has blocked a path traversal attack: path.normalize(...) originating from body.file.matches"
```
hansott, line 112: Why is this changed?

library/agent/protect.ts
```diff
@@ -132,6 +134,8 @@ function getWrappers() {
     new BetterSQLite3(),
     new Postgresjs(),
     new Fastify(),
+    new Koa(),
+    new KoaRouter(),
```
hansott, line 138: I wouldn't mind these being one sink but up to you!

library/sources/koa/wrapMiddleware.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable max-lines-per-function */
```
hansott, line 1:
```suggestion

```

library/sinks/Path.test.ts
timokoessler: Was changed in Node.js v21.1 😄

library/middleware/koa.ts
timokoessler: Ahh, yes 😅. Typescript always recommended this when trying to import with `import type { Application } from "koa"`.

library/agent/protect.ts
timokoessler: Because the koa router already consists of two different packages to wrap, it's maybe a bit better readable if it's splitted. Maybe there are more koa routers to add in the future.

hansott (approved):

PR #346, opened 2024-08-28
Enable SSRF redirect protection and add breaking test for AWS SDK v3

hansott:
* [ ] Check impact of #533 

(https://aikido-security.slack.com/archives/C06FA7ULFAQ/p1740684209786279)

PR #381, opened 2024-09-19
Only capture hostname when the request is not blocked because of SSRF

timokoessler:
This change also results in a different behavior when saving hostnames after a redirect using Undici / Fetch. Previously they were ignored, while redirects using the internal http or https module were already added to the list of outgoing hostnames.


library/vulnerabilities/ssrf/inspectDNSLookupCalls.ts
```diff
@@ -166,6 +175,9 @@ function wrapDNSLookupCallback(
     }
 
     if (!found) {
+      // Add to list of hostnames that the application is connecting to
+      agent.onConnectHostname(hostname, port);
```
hansott, line 179: Looking for a way to avoid the sprinkled `onConnectHostname`

Easy to forget when adding a new if statement here or there, right?

undici & fetch
```ts
const response = await fetch("url")
```
When response is returned and the fetch call didn't throw, we know the request succeeded, can we maybe hook into the return value and call `agent.onConnectHostname(hostname, port);` there?

http/https module

We could wrap the `on("response", (...) => {})` handler but we should NEVER add it when no response handler is (See #346)

So I think we have no choice to use the DNS lookup indeed, instead of adding the logic in `inspectDNSLookupCalls`, can we add a separate wrapper for this?

```ts
optionObj.lookup = inspectDNSLookupCalls(
      nativeLookup,
      agent,
      module,
      `${module}.request`,
      url,
      stackTraceCallingLocation
    ) as NonNullable<RequestOptions["lookup"]>;
```

becomes

```ts
optionObj.lookup = addHostnameToAgent(
     agent,
     inspectDNSLookupCalls(
      nativeLookup,
      agent,
      module,
      `${module}.request`,
      url,
      stackTraceCallingLocation
    )
);

function addHostnameToAgent(agent: Agent, lookup: Function) {
  return function addHostnameToAgent(...args: unknown[]) {
    // see if DNS lookup succeeded
    
    if (!error) {
      agent.onConnectHostname(hostname, port);
    }
  }
}
```

PR #386, opened 2024-09-19
Fix false positive with ASC/DESC

hansott:


hansott:
Will be fixed by https://github.com/AikidoSec/firewall-node/pull/401

PR #397, opened 2024-09-24
Check mongo distinct for NoSQL injection

timokoessler:


library/sinks/MongoDB.ts
```diff
@@ -137,17 +139,28 @@ export class MongoDB implements Wrapper {
...
+
+    if (filterPosition === 0 && args.length > 0 && isPlainObject(args[0])) {
+      filter = args[0];
+    } else if (
+      filterPosition === 1 &&
+      args.length > 1 &&
+      isPlainObject(args[1])
+    ) {
+      filter = args[1];
+    }
```
hansott, line 161: Maybe we can write this a bit simpler?

e.g. loop over all the args? And if it's a plainObject, inspect the filter?

timokoessler: Passed objects do not necessarily have to be a filter, they can also be e.g. options.

hansott (approved):

PR #404, opened 2024-10-02
Do not rate limit by ip if user is set

timokoessler:


library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -43,44 +43,45 @@ export function shouldRateLimitRequest(
...
+    updateContext(context, "consumedRateLimitForUser", true);
 
     if (!allowed) {
-      return { block: true, trigger: "ip" };
+      return { block: true, trigger: "user" };
     }
   }
 
-  if (context.user && !context.consumedRateLimitForUser) {
+  if (
```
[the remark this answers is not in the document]
timokoessler: The else block could also be triggered if `context.consumedRateLimitForUser` is set to true and a user is set.

library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -61,6 +61,14 @@ export function shouldRateLimitRequest(
         maxRequests
       );
 
+    if (context.remoteAddress && context.consumedRateLimitForIP) {
+      agent
+        .getRateLimiter()
+        .decrement(
+          `${endpoint.method}:${endpoint.route}:ip:${context.remoteAddress}`
+        );
```
hansott, line 69: Do we need to reset `consumedRateLimitForIP`? For correctness?

timokoessler: Currently it should be fine to do so, but if we allow resetting the user in the future, it could lead to unwanted behavior.

library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -21,6 +21,15 @@ export function shouldRateLimitRequest(
   context: Readonly<Context>,
   agent: Agent
 ): Result {
+  // Do not consume rate limit for the same request a second time
+  // (Might happen if the user adds the middleware multiple times)
+  if (context.consumedRateLimit) {
```
hansott, line 26: should we trigger a single time warning? like we did for setUser call?

timokoessler: Added

hansott (approved):

PR #408, opened 2024-10-07
Add express benchmark

timokoessler:
Simple empty route, compares requests per second

### DIFF
--- a/benchmarks/express/benchmark.js
+++ b/benchmarks/express/benchmark.js
@@ -1,0 +1,78 @@
+const { spawn } = require("child_process");
+const { setTimeout } = require("timers/promises");
+const { promisify } = require("util");
+const exec = promisify(require("child_process").exec);
+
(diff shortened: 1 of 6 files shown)


benchmarks/express/benchmark.js
```diff
@@ -0,0 +1,78 @@
+const { spawn } = require("child_process");
+const { setTimeout } = require("timers/promises");
+const { promisify } = require("util");
+const exec = promisify(require("child_process").exec);
+
+// Accepted percentage of performance decrease
+const AcceptedDecrease = 50; // %
```
hansott, line 7:
```suggestion
const AcceptedDecrease = 50; // %
```

Increase maybe?

timokoessler: Yes, thought it would be worse in GitHub Actions than locally.

hansott (approved):

PR #413, opened 2024-10-14
Add env var for max api discovery samples

timokoessler:
This pr adds a new environment variable `AIKIDO_MAX_API_DISCOVERY_SAMPLES` which defaults to `10` per route.


library/agent/api-discovery/updateApiInfo.ts
```diff
@@ -1,13 +1,23 @@
+import { getMaxApiDiscoverySamples } from "../../helpers/getMaxApiDiscoverySamples";
 import type { Context } from "../Context";
 import type { Route } from "../Routes";
 import { getApiInfo } from "./getApiInfo";
 import { mergeApiAuthTypes } from "./mergeApiAuthTypes";
 import { mergeDataSchemas } from "./mergeDataSchemas";
 
+const maxSamples = getMaxApiDiscoverySamples();
```
hansott, line 8: Maybe move this inside `updateApiInfo`?

I know it's unusual to change env vars at runtime but once this module is required, it will always stay the same value...

timokoessler: Moved. Many other env vars are only read at startup tho.

hansott (approved):

PR #417, opened 2024-10-15
Avoid double wrapping path

hansott:


library/sinks/Path.ts
```diff
@@ -34,19 +38,69 @@ export class Path implements Wrapper {
...
-      }
-    };
+    for (const func of functions) {
+      wrapExport(exports, func, pkgInfo, {
+        inspectArgs: (args) => this.inspectPath(args, func),
+      });
+    }
+  }
+
+  private isWindows() {
```
timokoessler, line 51: Maybe add as a global helper function?

timokoessler (approved):

PR #424, opened 2024-10-21
Log warning only once

hansott:


library/agent/context/user.ts
```diff
@@ -49,10 +49,8 @@ export function setUser(user: unknown) {
   }
 
   if (context.executedMiddleware) {
-    // eslint-disable-next-line no-console
-    console.warn(
-      `setUser(...) must be called before the Zen middleware is executed.`
-    );
+    logWarningSetUserCalledAfterMiddleware();
+    return;
```
timokoessler, line 53: Do we want to return here? Setting the user will still affect the users sent to core.

timokoessler (approved):

PR #432, opened 2024-10-28
Remove require() call from GraphQL source

timokoessler:
Adjustment of the behavior to the new hook system. Makes `--preserve-symlinks` no longer necessary in tests.

### DIFF
--- a/library/sources/GraphQL.ts
+++ b/library/sources/GraphQL.ts
@@ -12,6 +12,9 @@ import { shouldRateLimitOperation } from "./graphql/shouldRateLimitOperation";
 import { wrapExport } from "../agent/hooks/wrapExport";
 
 export class GraphQL implements Wrapper {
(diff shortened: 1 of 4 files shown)


library/sources/GraphQL.ts
```diff
@@ -119,6 +133,10 @@ export class GraphQL implements Wrapper {
       .withVersion("^16.0.0")
       .onFileRequire("execution/execute.js", (exports, pkgInfo) => {
         this.wrapExecution(exports, pkgInfo);
+      })
+      .onRequire((exports, pkgInfo) => {
+        this.GraphQLError = exports.GraphQLError;
+        this.visitFn = exports.visit;
```
hansott, line 139: Maybe combine them? And keep a reference to the original module? 

Then we can immediately return when execute is called and the module is not available for some reason?

PR #444, opened 2024-11-04
Perf opt: Check containsPrivateIPAddress(hostname) at start

hansott:
We don't need to iterate over the user input if the hostname is not a private IP address.


timokoessler (approved):

library/vulnerabilities/ssrf/checkContextForSSRF.ts
```diff
@@ -21,6 +21,10 @@ export function checkContextForSSRF({
   operation: string;
   context: Context;
 }): InterceptorResult {
+  // If the hostname is a private IP address, we don't need to check for SSRF
```
timokoessler: Missing `not` in comment?

PR #445, opened 2024-11-04
Don't add route or generate API spec if request comes from Aikido DAST

hansott:


library/agent/AikidoDAST.ts
```diff
@@ -0,0 +1,7 @@
+import { Context } from "./Context";
+
+const AIKIDO_DAST_HEADER = "aikido-api-test";
+
+export function isAikidoDASTRequest(context: Context) {
+  return context.headers[AIKIDO_DAST_HEADER] === "1";
```
timokoessler, line 6: Can `context.headers` be undefined here?

hansott: Added some extra checks, even though the types say headers will always be an object

PR #453, opened 2024-11-15
Make non-owned props of express wrapped functions accessible

hansott:


library/sources/Express.test.ts
```diff
@@ -554,3 +568,89 @@ t.test("it preserves original function name in Layer object", async () => {
...
+      throw new Error("server not found");
+    }
+
+    const address = server.address();
+
+    if (typeof address === "string") {
+      throw new Error("address is a string");
+    }
+
+    const response = await fetch(`http://localhost:${address!.port}/foo`);
```
timokoessler, line 642: Test fails in v16

library/sources/express/wrapRequestHandler.ts
```diff
@@ -1,37 +1,62 @@
...
- *
  * {
  *   value: 'myFunction',
  *   writable: false,
  *   enumerable: false,
  *   configurable: true
  * }
  */
-function preserveFunctionName(wrappedFunction: Function, originalName: string) {
+function preserveLayerName(wrappedFunction: Function, originalName: string) {
```
timokoessler, line 59: Would it make sense to preserve the function name in our wrap helpers in the future?

timokoessler (approved):

library/sources/Express.test.ts
hansott: Fixed, used our helper function to be able to run the test on v16!

PR #454, opened 2024-11-15
Report packages used at runtime

hansott:


library/agent/Agent.ts
```diff
@@ -36,6 +36,7 @@ export class Agent {
   private preventedPrototypePollution = false;
   private incompatiblePackages: Record<string, string> = {};
   private wrappedPackages: Record<string, WrappedPackage> = {};
+  private requiredPackages: Record<string, string[]> = {};
```
timokoessler, line 39: It doesn't make much difference, but perhaps we should use maps more often in the lib.
https://www.zhenghao.io/posts/object-vs-map

library/agent/hooks/wrapRequire.ts
```diff
@@ -188,11 +188,6 @@ function patchPackage(this: mod, id: string, originalExports: unknown) {
     .map((pkg) => pkg.getVersions())
     .flat();
 
-  // We don't want to patch this package because we do not have any hooks for it
-  if (!versionedPackages.length) {
-    return originalExports;
-  }
-
   // Read the package.json of the required package
   const packageJson = originalRequire(
```
timokoessler, line 192: Maybe we should wrap this in a try ... catch block then, as we run this for all modules now, so that Zen does not crash in case the package.json file does not exist for some unknown reason? (not sure if this is even possible)

library/agent/api/Event.ts
```diff
@@ -1,12 +1,15 @@
...
+type Version = string;
+type PackageName = string;
+
 export type AgentInfo = {
   dryMode: boolean;
   hostname: string;
   version: string;
   library: string;
-  packages: Record<string, string>;
+  packages: Record<PackageName, Version>;
```
bitterpanda63, line 12: Are we still going to use this one?

bitterpanda63: https://github.com/AikidoSec/firewall-node/blob/0d31cd93838f809026425ec913cedecd551b7134/library/agent/Agent.ts#L426-L436

library/agent/Agent.ts
hansott: It's using `Map` now 😊

library/agent/hooks/wrapRequire.ts
hansott: Added a try/catch to be safe!

library/agent/api/Event.ts
hansott: Let's keep it for now and remove in another PR?

library/agent/hooks/wrapRequire.ts
```diff
@@ -206,19 +206,24 @@ function patchPackage(this: mod, id: string, originalExports: unknown) {
     );
   }
 
+  const agent = getInstance();
+  agent?.onPackageRequired(moduleName, installedPkgVersion);
+
+  // We don't want to patch this package because we do not have any hooks for it
+  if (!versionedPackages.length) {
```
bitterpanda63, line 213: should we rename?

hansott: https://github.com/AikidoSec/firewall-node/pull/454/commits/8560275f272ca67e625904a646eb051541ad8fcc

PR #460, opened 2024-11-22
Support Prisma ORM

timokoessler:
Adds support for the ORM https://www.npmjs.com/package/prisma, because it's using rust db drivers.

Note: MongoDB is supported by Prisma only in replication mode, that's why a new container is added


library/sinks/Prisma.test.ts
```diff
@@ -0,0 +1,342 @@
...
+  );
+
+  if (stderr) {
+    t.fail(stderr);
+  }
+
+  // Clear require cache
+  for (const key in require.cache) {
+    delete require.cache[key];
+  }
```
hansott, line 169: why are we doing this? to be able to load prisma again? maybe cleaner to put in a different test file?

scripts/run-tap.js
```diff
@@ -11,11 +11,16 @@ if (process.env.CI) {
 }
 
 // Enable the `--experimental-sqlite` flag for Node.js ^22.5.0
-if (major === 22 && minor >= 5) {
+if ((major === 22 && minor >= 5) || major === 23) {
```
hansott, line 14: `isVersionGreaterOrEqual("22.5.0", getSemverNodeVersion())`?

library/agent/hooks/wrapExport.ts
```diff
@@ -154,7 +154,16 @@ function inspectArgs(
...
+  onInspectionInterceptorResult(context, agent, result, pkgInfo, start);
+}
 
+export function onInspectionInterceptorResult(
+  context: ReturnType<typeof getContext>,
+  agent: Agent,
+  result: InterceptorResult,
+  pkgInfo: WrapPackageInfo,
+  start: number
+) {
```
hansott, line 166: split because of length?

library/sinks/Prisma.test.ts
timokoessler: Yes, to load it again. Putting in different test files does not work because they are executed in parallel.

library/sinks/Prisma.ts
```diff
@@ -0,0 +1,247 @@
...
+          const agent = getInstance();
+          if (!agent) {
+            return;
+          }
+
+          // Extend all operations of the Prisma client
+          // https://www.prisma.io/docs/orm/prisma-client/client-extensions/query#modify-all-operations-in-all-models-of-your-schema
+          return instance.$extends({
+            query: {
+              $allOperations: ({
```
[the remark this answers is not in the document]
timokoessler: It's possible to use it multiple times, I can add a test for it 👍

scripts/run-tap.js
timokoessler: Can not import TS function from JS

hansott (approved):

PR #462, opened 2024-11-25
Fix Hono v4.6.9 and newer

timokoessler:


library/package.json
```diff
@@ -80,7 +80,7 @@
     "koa": "^2.15.3",
     "koa-router": "^12.0.1",
     "mariadb": "^3.3.2",
-    "mongodb": "^6.3.0",
+    "mongodb": "6.9",
```
hansott, line 83:
```suggestion
    "mongodb": "~6.9",
```

library/sources/Hono.ts
```diff
@@ -28,15 +41,21 @@ export class Hono implements Wrapper {
...
-        wrapExport(exports.HonoBase.prototype, "addRoute", pkgInfo, {
-          modifyArgs: (args) => this.wrapArgs(args),
-        });
-      })
-      .onFileRequire("dist/cjs/hono-base.js", (exports, pkgInfo) => {
-        wrapExport(exports.HonoBase.prototype, "addRoute", pkgInfo, {
-          modifyArgs: (args) => this.wrapArgs(args),
-        });
+      .onRequire((exports, pkgInfo) => {
+        return {
```
hansott, line 45: Can we still include the other exports somehow? What if they add a new property?

timokoessler: Changed like you suggested.

PR #463, opened 2024-11-25
Mark JS values as unsafe

hansott:
e.g. output from LLM that is parsed to JSON


library/agent/context/markUnsafe.ts
```diff
@@ -0,0 +1,32 @@
...
+  const agent = getInstance();
+
+  if (!agent) {
+    return;
+  }
+
+  const context = ContextStorage.getStore();
+
+  if (!context) {
+    return;
```
timokoessler, line 15: If the user calls this method outside of the context of an incoming http request, he might never notice that it doesn't have any effect to call the method?

library/agent/context/markUnsafe.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const current = context.markUnsafe || [];
+
+  const a = JSON.stringify(payload);
+
+  if (
+    !current.some((item) => {
+      // JSON.stringify is used to compare objects
+      // without having to copy a deep equality function
+      return JSON.stringify(item) === a;
```
hansott, line 26: todo: add try/catch

library/agent/context/markUnsafe.ts
```diff
@@ -0,0 +1,32 @@
+import { getInstance } from "../AgentSingleton";
+import { updateContext } from "../Context";
+import { ContextStorage } from "./ContextStorage";
+
+export function markUnsafe(payload: unknown) {
```
hansott, line 5: todo: add docs

library/agent/context/markUnsafe.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!context) {
+    logWarningMarkUnsafeWithoutContext();
+    return;
+  }
+
+  try {
+    const current = context.markUnsafe || [];
+    const a = JSON.stringify(payload);
+    console.log("a", a);
```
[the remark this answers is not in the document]
hansott: Fixed!!

timokoessler (approved):

timokoessler (approved):

PR #471, opened 2024-12-05
Use process.nextTick before wrapping Collection (mongodb v6.10.0)

hansott:
```js
exports.Collection = void 0;
const collection_1 = require("./collection");
Object.defineProperty(exports, "Collection", { enumerable: true, get: function () { return collection_1.Collection; } });
```

For some reason the Collection is undefined in our callback, let's wait until the next tick to grab a reference to Collection.

### DIFF
--- a/library/sinks/MongoDB.ts
+++ b/library/sinks/MongoDB.ts
@@ -2,6 +2,7 @@
 import type { Collection } from "mongodb-v6";
 import { Hooks } from "../agent/hooks/Hooks";
 import { InterceptorResult } from "../agent/hooks/InterceptorResult";
+import type { WrapPackageInfo } from "../agent/hooks/WrapPackageInfo";
 import { detectNoSQLInjection } from "../vulnerabilities/nosql-injection/detectNoSQLInjection";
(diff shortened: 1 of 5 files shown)


library/sinks/MongoDB.ts
```diff
@@ -186,33 +187,48 @@ export class MongoDB implements Wrapper {
...
-        wrapExport(collectionProto, "distinct", pkgInfo, {
-          inspectArgs: (args, agent, collection) =>
-            this.inspectDistinct(args, collection as Collection),
+        // From mongodb v6.10.0, the Collection is undefined
+        // It's defined like:
+        // exports.Collection = void 0;
+        // const collection_1 = require("./collection");
+        // Object.defineProperty(exports, "Collection", { enumerable: true, get: function () { return collection_1.Collection; } });
+        // So we need to wait for the next tick to wrap the Collection
+        process.nextTick(() => {
```
timokoessler, line 230: We could check if the export is undefined. If not, we don't need to wait one tick.

timokoessler (approved):

PR #478, opened 2024-12-17
Add support for bot blocking

timokoessler:


library/agent/ServiceConfig.ts
```diff
@@ -127,6 +128,21 @@ export class ServiceConfig {
     this.setBlockedIPAddresses(blockedIPAddresses);
   }
 
+  updateBlockedUserAgents(blockedUserAgents: string) {
+    if (!blockedUserAgents) {
+      this.blockedUserAgentRegex = undefined;
+      return;
+    }
+    this.blockedUserAgentRegex = new RegExp(blockedUserAgents, "i");
```
hansott, line 136: This might throw?

timokoessler: The location where this method is called is already surrounded by a try/catch.

hansott (approved):

PR #488, opened 2024-12-24
Prevent crash and counting too much if there are multiple request handlers

hansott:


library/sources/http-server/createRequestListener.ts
```diff
@@ -68,8 +71,24 @@ function callListenerWithContext(
   });
 }
 
-function createOnFinishRequestHandler(res: ServerResponse, agent: Agent) {
+// Use symbol to avoid conflicts with other properties
+const countedRequest = Symbol("__zen__");
```
timokoessler, line 75: Maybe also use a symbol name that describes the usage and not only zen?

PR #505, opened 2025-01-21
Report stats about blocked and rate limited requests

hansott:


timokoessler (approved):

library/agent/InspectionStatistics.test.ts
```diff
@@ -462,6 +537,54 @@ t.test("it keeps track of aborted requests", async () => {
...
+    match: "ipBlocklist",
+    key: "known_threat_actors/public_scanners",
+  });
+  stats.onBlockedRequest({ match: "userAgentList", key: "ai_data_scrapers" });
+
+  t.same(stats.getStats(), {
+    sinks: {},
+    startedAt: 0,
+    requests: {
+      total: 0,
```
timokoessler, line 569: Blocked requests are not counted as total requests?

hansott:
```js
agent.getInspectionStatistics().onRequest();
```

is called in `createOnFinishRequestHandler`, that one will increment the total requests

Maybe we can refactor this later so that we have a single method to increment the counts together?

timokoessler (approved):

library/agent/ServiceConfig.ts
```diff
@@ -2,23 +2,34 @@ import { IPMatcher } from "../helpers/ip-matcher/IPMatcher";
...
+  private blockedIPAddresses: {
+    key: string;
+    blocklist: IPMatcher;
+    description: string;
+  }[] = [];
   private blockedUserAgentRegex: RegExp | undefined;
+  private monitoredUserAgents: {
+    key: string;
+    pattern: RegExp;
+  }[] = [];
```
timokoessler, line 22: No strong opinion about that, but do we even need two different lists? We could add a property to the existing list entries. Or are other agents validating additional keys?

library/agent/InspectionStatistics.ts
```diff
@@ -176,6 +226,50 @@ export class InspectionStatistics {
     }
   }
 
+  onIPAddressMatches(matches: { key: IPListKey; monitor: boolean }[]) {
+    if (matches.length > 0) {
+      this.ipAddresses.total += 1;
+    }
+
+    const blockingMatches = matches.filter((match) => !match.monitor);
```
timokoessler, line 234: Performance: Filter will loop over all elements, we could break on first match

library/agent/api/fetchBlockedLists.ts
```diff
@@ -3,23 +3,35 @@ import { getAPIURL } from "../getAPIURL";
...
 export type IPList = {
+  key: string;
   source: string;
   description: string;
   ips: string[];
+  monitor: boolean;
 };
 
-export async function fetchBlockedLists(token: Token): Promise<{
+export type AgentBlockList = {
```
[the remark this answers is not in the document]
hansott: Renamed to `BotBlocklist`

library/sources/http-server/checkIfRequestIsBlocked.ts
```diff
@@ -65,38 +65,47 @@ export function checkIfRequestIsBlocked(
     return true;
   }
 
-  const result = context.remoteAddress
+  const blockedIPs = context.remoteAddress
     ? agent.getConfig().isIPAddressBlocked(context.remoteAddress)
```
timokoessler, line 69: A bit confusing that the method `isIPAddressBlocked` can also return a non-empty array if the ip is not blocked (monitor mode). Same for user agents. Maybe rename method?

library/agent/InspectionStatistics.ts
timokoessler:
```suggestion
    const blockingMatch = matches.find(match) => !match.monitor);
```

Multiline suggestion does not work 😢

hansott: See https://github.com/AikidoSec/firewall-node/pull/505/commits/2fa9007447fd85386d50db632c3677ce800c294e

timokoessler (approved):

PR #510, opened 2025-01-27
Show connection failures on startup

timokoessler:
Extracted from #435

### DIFF
--- a/library/agent/Agent.ts
+++ b/library/agent/Agent.ts
@@ -1,4 +1,4 @@
-/* eslint-disable max-lines-per-function */
+/* eslint-disable max-lines-per-function, no-console */
 import { hostname, platform, release } from "os";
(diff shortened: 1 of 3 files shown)


library/agent/Agent.ts
```diff
@@ -193,8 +205,10 @@ export class Agent {
     this.attackLogger.log(attack);
 
     if (this.token) {
-      this.api.report(this.token, attack, this.timeoutInMS).catch(() => {
-        this.logger.log("Failed to report attack");
+      this.api.report(this.token, attack, this.timeoutInMS).catch((err) => {
+        console.error(
+          `Aikido: Failed to report attack event to Aikido platform: ${err.message}`
+        );
```
hansott, line 211: What if token was rotated when the app is running?

hansott (approved):

PR #523, opened 2025-02-10
Simplify and comment shouldDiscoverRoute

timokoessler:


library/sources/http-server/shouldDiscoverRoute.ts
```diff
@@ -1,7 +1,7 @@
-import { extname } from "path";
+import { getFileExtension } from "../../helpers/getFileExtension";
 
 const EXCLUDED_METHODS = ["OPTIONS", "HEAD"];
-const IGNORE_EXTENSIONS = ["properties", "php", "asp", "aspx", "jsp", "config"];
+const IGNORE_EXTENSIONS = ["properties", "config"];
```
bitterpanda63, line 4:
```suggestion
// These are extensions that are too long to get ignored, we manually ignore them
const IGNORE_EXTENSIONS = ["properties", "config"];
```
Or something like that

library/sources/http-server/shouldDiscoverRoute.ts
```diff
@@ -34,23 +34,27 @@ export function shouldDiscoverRoute({
     return false;
   }
 
-  return segments.every(isAllowedExtension);
+  // Check for every file segment if it contains an file extension and if it should be discovered or ignored
+  return segments.every(shouldDiscoverExtension);
```
hansott, line 38: We only need to do this for the full string right? Not for every segment, maybe we can refactor?

timokoessler: Required for `/index.php/something`

hansott (approved):

PR #526, opened 2025-02-10
Log warning if used with ESM

timokoessler:


hansott (approved):

library/index.ts
```diff
@@ -8,11 +8,18 @@ import { addHonoMiddleware } from "./middleware/hono";
...
 import { addKoaMiddleware } from "./middleware/koa";
+import { isESM } from "./helpers/isESM";
 
 const supported = isFirewallSupported();
 const shouldEnable = shouldEnableFirewall();
 
 if (supported && shouldEnable) {
+  if (isESM()) {
+    console.warn(
+      "AIKIDO: Your application seems to be running in ESM mode. Zen does not support ESM at runtime yet."
```
willem-delbare, line 19: why don't we return then?

timokoessler: It just feels safer in case there are false positive detections of the execution mode. One possibility I can think of but that's really rare is a ESM module loaded in a CJS project using something like TypeScript Execute (tsx) that also wraps require before we do.

PR #536, opened 2025-02-21
Improve SSRF performance: re-use extracted hostname from URL

hansott:


library/vulnerabilities/ssrf/findHostnameInUserInput.ts
```diff
@@ -1,29 +1,26 @@
...
   port?: number
 ): boolean {
   if (userInput.length <= 1) {
     return false;
   }
 
-  const hostnameURL = tryParseURL(`http://${hostname}`);
-  if (!hostnameURL) {
-    return false;
-  }
```
hansott, line 16: This currently happens for every user input

PR #540, opened 2025-02-27
New instrumentation system

timokoessler:


instrumentation-wasm/src/js_transformer/helpers/select_sourcetype_based_on_enum.rs
```diff
@@ -0,0 +1,21 @@
...
+*/
+pub fn select_sourcetype_based_on_enum(src_type: i32) -> SourceType {
+    // 0 is generic type.
+    match src_type {
+        0 => SourceType::unambiguous(),
+        1 => SourceType::ts(),
+        2 => SourceType::cjs(),
+        3 => SourceType::mjs(),
+        4 => SourceType::tsx(),
+        _ => SourceType::unambiguous(),
```
hansott, line 12: maybe we can also pass as a string inside the instructions? We're already parsing JSON?

Easier to see in the JS side?

I know we did this for Zen internals but speed and memory is more important there?

library/agent/hooks/instrumentation/injectedFunctions.ts
```diff
@@ -0,0 +1,84 @@
...
+    name: id,
+    type: "builtin",
+  };
+
+  // Todo check if cache is needed
+  for (const interceptor of interceptors) {
+    try {
+      const returnVal = interceptor(exports, pkgInfo);
+      // If the interceptor returns a value, we want to use this value as the new exports
+      if (typeof returnVal !== "undefined") {
```
hansott:
```suggestion
      if (returnVal !== undefined) {
```

instrumentation-wasm/src/js_transformer/helpers/parse_js_code_to_statements.rs
```diff
@@ -0,0 +1,19 @@
...
+use oxc_span::SourceType;
+
+pub fn parse_js_code_to_statements<'a>(
+    allocator: &'a Allocator,
+    source_text: &'a str,
+    source_type: SourceType,
+) -> Vec<'a, Statement<'a>> {
+    let parser_result = Parser::new(&allocator, source_text, source_type)
+        .with_options(ParseOptions {
+            allow_return_outside_function: true,
```
hansott: Might be nice to add a comment for this param?

scripts/build.js
```diff
@@ -102,6 +126,22 @@ async function modifyDtsFilesAfterBuild() {
...
 }
 
+async function buildInstrumentationWasm() {
+  // Build Instrumentation WASM
+  await execAsync(
+    `wasm-pack build --release --target nodejs --out-dir ${instrumentationWasmOutDir}`,
+    {
+      cwd: instrumentationWasmDir,
+    }
+  );
```
hansott, line 157: Can we output what the command is doing maybe? It seems to hang but it's compiling rust of course :D 

`stdio: 'inherit`?

instrumentation-wasm/src/js_transformer/helpers/get_method_arg_names.rs
```diff
@@ -0,0 +1,16 @@
...
+    method_definition
+        .value
+        .params
+        .items
+        .iter()
+        .for_each(|param| {
+            arg_names.push(param.pattern.get_identifier_name().unwrap().to_string());
+        });
+
+    arg_names
```
hansott: What about spread arguments?

instrumentation-wasm/src/js_transformer/helpers/get_import_code_str.rs
```diff
@@ -0,0 +1,10 @@
+use oxc_span::SourceType;
+
+pub fn get_import_code_str(source_type: &SourceType, has_module_syntax: bool) -> &'static str {
+    // Todo maybe import only needed functions, not sure if it is worth it
+    if source_type.is_script() || source_type.is_unambiguous() && !has_module_syntax {
```
hansott: x || y && !z

difficult to reason about :D

instrumentation-wasm/src/js_transformer/helpers/parse_js_code_to_statements.rs
```diff
@@ -0,0 +1,19 @@
...
+    source_type: SourceType,
+) -> Vec<'a, Statement<'a>> {
+    let parser_result = Parser::new(&allocator, source_text, source_type)
+        .with_options(ParseOptions {
+            allow_return_outside_function: true, // Could be partical code string that should be parsed
+            ..ParseOptions::default()
+        })
+        .parse();
+
+    return parser_result.program.body;
```
hansott: Check panic flag? https://github.com/oxc-project/oxc/blob/main/crates/oxc_parser/src/lib.rs#L179

And whether there are any errors?

Return maybe/option?

library/agent/hooks/instrumentation/codeTransformation.ts
```diff
@@ -0,0 +1,35 @@
...
+): string {
+  const result = wasm_transform_code_str(
+    pkgName,
+    pkgVersion,
+    code,
+    JSON.stringify(fileInstructions),
+    getSourceType(path, pkgLoadFormat)
+  );
+
+  if (result.startsWith("#ERR:")) {
```
hansott: ehhh, there's no nicer way to throw/catch errors? :D

library/agent/hooks/instrumentation/getExportsForBuiltin.ts
```diff
@@ -0,0 +1,9 @@
+export function getExportsForBuiltin(name: string) {
+  const mod = process.getBuiltinModule(name);
+
+  if (!mod) {
+    return undefined;
+  }
+
+  return new Set(["default", ...Object.keys(mod)]);
```
hansott: Probably can use some comments here :D

library/agent/hooks/instrumentation/getPackageVersionFromPath.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  try {
+    const version = JSON.parse(
+      readFileSync(`${basePath}/package.json`, "utf8")
+    ).version;
+    cache.set(basePath, version);
+    return version;
+  } catch {
+    // Return undefined if the package is not found
```
hansott, line 23: it will try this every time? should we cache it? e.g. `__NOT_FOUND__` or something like that?

instrumentation-wasm/src/js_transformer/helpers/get_name_str_for_member_expr.rs
```diff
@@ -0,0 +1,48 @@
+use oxc_allocator::{Allocator, CloneIn};
+use oxc_ast::ast::{Expression, MemberExpression};
+
+pub fn get_name_str_for_member_expr<'a>(
```
hansott, line 11: maybe put example in a comment (in JavaScript)?

instrumentation-wasm/src/js_transformer/helpers/insert_code.rs
```diff
@@ -0,0 +1,203 @@
...
+    // Build and add a call expression
+    let call_expr = builder.expression_call(
+        SPAN,
+        builder.expression_identifier(SPAN, "__instrumentInspectArgs"),
+        NONE,
+        inspect_args,
+        false,
+    );
+
+    let stmt_expression = builder.statement_expression(SPAN, call_expr);
```
hansott, line 51: Feels like it's easier to build a JS string and then parse and add to AST? Just a thought, no need to change (there's perhaps something I'm missing)

instrumentation-wasm/src/js_transformer/helpers/insert_import_statement.rs
```diff
@@ -0,0 +1,139 @@
...
+pub fn insert_import_statement<'a>(
+    source_type: &SourceType,
+    has_module_syntax: bool,
+    allocator: &'a Allocator,
+    builder: &'a AstBuilder,
+    body: &mut OxcVec<'a, Statement<'a>>,
+    file_instructions: &FileInstructions,
+) {
+    // Common JS require() statement
+    if source_type.is_script() || source_type.is_unambiguous() && !has_module_syntax {
```
hansott: Same here, this boolean logic breaks my brain :D

library/agent/Agent.ts
```diff
@@ -68,11 +69,16 @@ export class Agent {
     private readonly logger: Logger,
     private readonly api: ReportingAPI,
     private readonly token: Token | undefined,
-    private readonly serverless: string | undefined
+    private readonly serverless: string | undefined,
+    private readonly newInstrumentation: boolean = false
```
hansott: Do we ever pass this in the constructor? Otherwise put as a normal property?

library/agent/Agent.ts
```diff
@@ -68,11 +69,16 @@ export class Agent {
...
   ) {
     if (typeof this.serverless === "string" && this.serverless.length === 0) {
       throw new Error("Serverless cannot be an empty string");
     }
+
+    if (envToBool(process.env.AIKIDO_TEST_NEW_INSTRUMENTATION)) {
```
hansott: Maybe we can extract a function for this? not sure

library/agent/hooks/instrumentation/index.ts
```diff
@@ -0,0 +1,29 @@
...
+let hooksRegistered = false;
+
+export function registerNodeHooks() {
+  if (!("registerHooks" in mod) || typeof mod.registerHooks !== "function") {
+    throw new Error("This Node.js version is not supported");
+  }
+
+  // Do not register hooks multiple times
+  if (hooksRegistered) {
+    return;
```
hansott, line 16: Do we need to log something? Is it possible that a user could trigger this?

library/agent/hooks/instrumentation/index.ts
```diff
@@ -0,0 +1,29 @@
+import { onModuleLoad } from "./loadHook";
+import * as mod from "node:module";
+import type { RegisterHookFunction } from "./types";
+import { patchProcessGetBuiltinModule } from "./processGetBuiltin";
+
+let hooksRegistered = false;
+
+export function registerNodeHooks() {
+  if (!("registerHooks" in mod) || typeof mod.registerHooks !== "function") {
+    throw new Error("This Node.js version is not supported");
```
hansott, line 11: State from which version it's supported?

library/package.json
```diff
@@ -125,6 +135,7 @@
   "scripts": {
     "test": "node ../scripts/run-tap.js",
     "test:ci": "CI=true node ../scripts/run-tap.js",
+    "test:new": "node ../scripts/run-tap.js --test-new-instrumentation",
```
hansott, line 161: let's add a list of tests that are passing right now? So that we can run them as part of CI? but the remaining ones should be allowed to fail?

instrumentation-wasm/src/js_transformer/helpers/parse_js_code_to_statements.rs
timokoessler: File does not exist anymore

instrumentation-wasm/src/js_transformer/helpers/parse_js_code_to_statements.rs
timokoessler: Function removed

library/agent/hooks/instrumentation/index.ts
timokoessler: Only if two separate agents are started without using the main library exports, so only during unit tests.

instrumentation-wasm/src/js_transformer/helpers/insert_code.rs
timokoessler: Easier to write indeed, but also slower and more unsafe.

instrumentation-wasm/src/js_transformer/helpers/select_sourcetype_based_on_enum.rs
```diff
@@ -0,0 +1,21 @@
...
+Default -> JS, auto-detect CJS or ESM
+*/
+pub fn select_sourcetype_based_on_enum(src_type: i32) -> SourceType {
+    // 0 is generic type.
+    match src_type {
+        0 => SourceType::unambiguous(),
+        1 => SourceType::ts(),
+        2 => SourceType::cjs(),
+        3 => SourceType::mjs(),
+        4 => SourceType::tsx(),
```
timokoessler: Missing jsx

library/agent/hooks/instrumentation/getSourceType.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+    case "cjs": {
+      return 2;
+    }
+    case "mjs": {
+      return 3;
+    }
+    case "tsx": {
+      return 4;
+    }
```
timokoessler, line 30: JSX is missing

library/agent/hooks/instrumentation/getExportsForBuiltin.ts
timokoessler: Not used anymore, removing

instrumentation-wasm/src/js_transformer/helpers/get_method_arg_names.rs
timokoessler: They are currently not supported, but simply ignored. There is a unit test for this in `codeTransformation.test.ts`. Alternatively `modifyArgumentsObject = true` could be used.

docs/esm.md
```diff
@@ -0,0 +1,44 @@
...
+
+By default, the firewall will run in non-blocking mode. When it detects an attack, the attack will be reported to Aikido if the environment variable `AIKIDO_TOKEN` is set and continue executing the call.
+
+You can enable blocking mode by setting the environment variable `AIKIDO_BLOCK` to `true`:
+
+```sh
+AIKIDO_BLOCK=true node app.js
+```
+
+It's recommended to enable this on your staging environment for a considerable amount of time before enabling it on your production environment (e.g. one week).
```
hansott: Do we need this bit? Seems unrelated to ESM

scripts/run-tap.js
```diff
@@ -15,12 +14,44 @@ if (process.env.CI) {
...
+
+  // This list excludes test files of sinks and sources that are not yet ported to the new instrumentation
+  const excludedTestFilesForNewInstrumentation = [
+    "**/sinks/Undici*",
+    "**/sinks/Prisma.test.ts",
+    "**/sinks/AwsSDK*",
+    "**/sinks/AiSDK.test.ts",
+    "**/sinks/Anthropic.test.ts",
+    "**/sinks/OpenAI.test.ts",
+    // Todo support AWS Bedrock
```
hansott: AwsSDK is excluded? So we can remove the todo comment as this list is our todo list? :D 

```suggestion
```

library/sources/PubSub.ts
```diff
@@ -5,27 +5,53 @@ import { Wrapper } from "../agent/Wrapper";
...
+            nodeType: "MethodDefinition",
+            operationKind: undefined,
+            inspectArgs: (args, agent, subscription) => {
+              wrapExport(
+                subscription,
+                "on",
+                {
+                  name: "@google-cloud/pubsub",
+                  type: "external",
+                },
```
hansott, line 46: I'm fine with specifying the package info statically (`WrapPackageInfo`)
I'm not a fan of adding a callback to `addFileInstrumentation` or adding another argument to `inspectArgs`

However, the package info is used in `wrapExport`, in theory the `WrapPackageInfo` should have a path if it's an external module?

We should improve the type so that it never can be in an invalid state:

```ts
export type WrapPackageInfo =
  | {
      // Name of the builtin module
      name: string;
      type: "builtin";
    }
  | {
      // Name of the external package
      name: string;
      // Version of the external package
      version: string;
      type: "external";
      // Path information for the external package
      path: {
        base: string;
        // Path of the imported js file relative to the module base directory
        relative: string;
      };
    }
  | {
      // Name of the global
      name: string;
      type: "global";
    };
```

What's also an option, accept a reduced package info type:

```ts
type WrapPackageInfo = { name: string; type: "global" | "external" | "builtin" }
```

library/sources/Fastify.ts
```diff
@@ -144,6 +145,42 @@ export class Fastify implements Wrapper {
...
+          }),
+      },
+      {
+        name: "_route",
+        nodeType: "FunctionExpression",
+        operationKind: undefined,
+        modifyArgs: this.wrapRouteMethod,
+      },
+      ...requestFunctions.map(
+        (func) =>
```
hansott:
```suggestion
        (func): PackageFunctionInstrumentationInstruction =>
```

Instead of `as PackageFunctionInstrumentationInstruction`?

library/sinks/MongoDB.ts
```diff
@@ -233,6 +234,46 @@ export class MongoDB implements Wrapper {
...
+                name: operation,
+                nodeType: "MethodDefinition",
+                operationKind: "nosql_op",
+                inspectArgs: (args, agent, collection) =>
+                  this.inspectOperation(
+                    operation,
+                    args,
+                    collection as Collection
+                  ),
+              }) as PackageFunctionInstrumentationInstruction
```
hansott: see other comment!

sample-apps/hono-pg-esm/app.js
```diff
@@ -0,0 +1,66 @@
...
+import { createConnection } from "./db.js";
+
+const app = new Hono();
+const db = await createConnection();
+
+// Check until we have ESM unit tests
+if (!readFile.__wrapped || !fs.readFile.__wrapped) {
+  console.error("fs.readFile is not wrapped");
+  process.exit(1);
+}
```
hansott, line 15: We still need this because we don't run our tests in ESM mode, right?

timokoessler: Yes, still have to find a good stable solution for that.

docs/esm.md
```diff
@@ -0,0 +1,32 @@
...
+export NODE_OPTIONS='-r @aikidosec/firewall/instrument'
+```
+
+> [!IMPORTANT]  
+> Please also check the documentation on how to integrate Zen with your used web framework.
+
+## Known issues
+
+- The app might crash on startup if used together with some packages that use the Node.js Asynchronous Module Customization Hooks, like Sentry or OpenTelemetry, due to bugs in Node.js itself.
+- Zen can not protect ESM sub-dependencies of a ESM package. For example if a ESM package `foo` imports a sub-dependency `bar` that is also an ESM package, Zen will not be able to protect the code in `bar`. This is because the V8 engine does not allow Node.js to observe the evaluation of inner ESM packages (yet).
```
hansott:
```suggestion
- Zen can not protect ESM sub-dependencies of an ESM package. For example if an ESM package `foo` imports a sub-dependency `bar` that is also an ESM package, Zen will not be able to protect the code in `bar`. This is because the V8 engine does not allow Node.js to observe the evaluation of inner ESM packages (yet).
```

docs/esm.md
```diff
@@ -0,0 +1,32 @@
...
+
+Relevant links:
+
+- [ERR_INVALID_RETURN_PROPERTY_VALUE when using module.register and module.registerHooks (#57327)](https://github.com/nodejs/node/issues/57327)
+- [module.registerHooks() tracking issue (#56241)](https://github.com/nodejs/node/issues/56241)
+- [TypeError when json file is required in hook and in the imported file (#57358)](https://github.com/nodejs/node/issues/57358)
+- [ERR_INTERNAL_ASSERTION: Unexpected module status 3 (#58515)](https://github.com/nodejs/node/issues/58515)
+- [Adding an evaluation hook for v8::Module](https://issues.chromium.org/u/1/issues/384413088)
```
hansott: Let's add this at the end?

### Example of unprotected ESM sub-dependency

Consider this scenario where your application uses an ESM package that has ESM sub-dependencies:

**Your app.js:**
```js
import { logUserAction } from 'analytics-client'; // ESM package

app.post('/user/action', async (req, res) => {
  await logUserAction(req.body.userId, req.body.action);
  res.json({ success: true });
});
```

**node_modules/analytics-client/index.js (dependency):**
```js
import { storeEvent } from 'analytics-db-helper';

export async function logUserAction(userId, action) {
  const eventData = `INSERT INTO user_events (user_id, action, timestamp) VALUES (${userId}, '${action}', NOW())`;
  return await storeEvent(eventData);
}
```

**node_modules/analytics-db-helper/index.js (another sub-dependency):**
```js
import mysql from 'mysql2/promise';

export async function storeEvent(sql) {
  // Zen CANNOT instrument this mysql.execute call!
  // V8 doesn't allow Node.js to observe evaluation of this inner ESM module
  const connection = await mysql.createConnection(config);
  try {
    const result = await connection.execute(sql);
    return result;
  } finally {
    await connection.end();
  }
}
```

In this example, if `req.body` contained a SQL injection payload, Zen would miss detecting it at the actual `mysql.execute` level inside the ESM sub-dependency, since it cannot instrument third-party packages within ESM sub-dependencies.

PR #541, opened 2025-02-28
Ignore .well-known routes if not actual well known URI

hansott:


library/sources/http-server/shouldDiscoverRoute.ts
```diff
@@ -61,11 +63,6 @@ function shouldDiscoverExtension(segment: string) {
 }
 
 function isDotFile(segment: string) {
-  // See https://www.rfc-editor.org/rfc/rfc8615
-  if (segment === ".well-known") {
```
bitterpanda63, line 65: This check can stay no? We want to ignore the other .well-known files still right ("fake" .well-known files)

bitterpanda63: I guess this check was always unnecessary haha

bitterpanda63 (approved):

library/helpers/isWellKnownURI.ts
```diff
@@ -0,0 +1,98 @@
...
+  "/.well-known/webfinger",
+  "/.well-known/webweaver.json",
+  "/.well-known/wot",
+];
+
+// Check if a path is a well-known URI
+// e.g. /.well-known/acme-challenge
+// https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml
+export function isWellKnownURI(path: string) {
+  return wellKnown.includes(path);
```
timokoessler, line 97: A Set might have a better performance?

timokoessler (approved):

PR #549, opened 2025-03-14
Fix false positive SSRF with HTTP/HTTPS

hansott:


library/vulnerabilities/ssrf/checkContextForSSRF.ts
```diff
@@ -40,12 +40,19 @@ export function checkContextForSSRF({
...
-    ) {
-      return;
+    if (baseURL && baseURL.hostname === hostname) {
+      const baseURLPort = getPortFromURL(baseURL);
+      if (baseURLPort === port) {
+        return undefined;
+      }
+      // Special case for HTTP/HTTPS ports
+      // In production, the app will be served on port 80 and 443
+      if (baseURLPort === 80 && port === 443) {
```
bitterpanda63, line 50: would put this code all in one place right, if we were to change, now its 2x same code

hansott: Good idea, thanks!

library/vulnerabilities/ssrf/checkContextForSSRF.ts
```diff
@@ -31,22 +32,20 @@ export function checkContextForSSRF({
   // This is just to inspect direct invocations of `http.request` and similar
   // Where the hostname might be a private IP address (or localhost)
   if (!containsPrivateIPAddress(hostname)) {
-    return;
+    return undefined;
```
bitterpanda63, line 35: is related to pr?

library/vulnerabilities/ssrf/isRequestToItself.ts
```diff
@@ -0,0 +1,49 @@
...
+export function isRequestToItself({
+  serverUrl,
+  outboundHostname,
+  outboundPort,
+}: {
+  serverUrl: string;
+  outboundHostname: string;
+  outboundPort: number | undefined;
+}): boolean {
+  // If Node.js is exposed to the internet, we can't be sure about the Host header
```
bitterpanda63, line 17: comment is a bit confusing to read, talking about exposed to the internet when using x-forwarded-for and "Host" header only. possible rewrite? The logic is smart though

bitterpanda63 (approved):

PR #556, opened 2025-03-20
Add kind to sink stats

hansott:
To know what kind of operations the sink does


timokoessler (approved):

library/sources/FastXmlParser.ts
```diff
@@ -45,6 +45,7 @@ export class FastXmlParser implements Wrapper {
       .onRequire((exports, pkgInfo) => {
         wrapNewInstance(exports, "XMLParser", pkgInfo, (instance) => {
           wrapExport(instance, "parse", pkgInfo, {
+            kind: "unserialize_op",
```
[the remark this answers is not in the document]
hansott: yeah, you're right! but can't hurt to have it? if we would use `inspectArgs` later?

library/agent/api/Event.ts
```diff
@@ -64,7 +64,19 @@ export type DetectedAttack = {
...
 
-type MonitoredSinkStats = {
+export type OperationKind =
+  | "sql_op"
+  | "nosql_op"
+  | "outgoing_http_op"
+  | "fs_op"
+  | "path_op"
+  | "exec_op"
+  | "unserialize_op"
```
bitterpanda63, line 74: would call deserialize

library/agent/api/Event.ts
```diff
@@ -64,7 +64,19 @@ export type DetectedAttack = {
   time: number;
 };
 
-type MonitoredSinkStats = {
+export type OperationKind =
+  | "sql_op"
+  | "nosql_op"
+  | "outgoing_http_op"
+  | "fs_op"
```
bitterpanda63, line 71: would merge fs_op and path_op, the difference is not as clear in every language

library/agent/api/Event.ts
```diff
@@ -82,7 +94,7 @@ type MonitoredSinkStats = {
 type Heartbeat = {
   type: "heartbeat";
   stats: {
-    sinks: Record<string, MonitoredSinkStats>;
+    operations: OperationStats[];
```
bitterpanda63, line 97: would still provide a key no? interesting data

library/agent/api/Event.ts
hansott: yes, makes sense! I also thought about that

library/agent/api/Event.ts
hansott: previously we used the sink name (e.g. mongodb) as key for the object

I moved the `kind` inside the array now, I think it makes it easier to change the format later on 😬

PR #564, opened 2025-03-27
Support ip ranges in endpoint allowedIPAddresses

timokoessler:


library/agent/Config.ts
```diff
@@ -14,8 +16,12 @@ export type Endpoint = {
   };
 };
 
+export type Endpoint = Omit<EndpointConfig, "allowedIPAddresses"> & {
+  allowedIPAddresses?: IPMatcher;
```
hansott, line 20: Took me a second to know why we use `Omit` but otherwise `allowedIPAddresses` becomes `never`, right?

```suggestion
  allowedIPAddresses: IPMatcher | undefined;
```

I think I might prefer setting it explicitly to `undefined` (now the property key doesn't need to exist)?

timokoessler: The issue is that it still allows `allowedIPAddresses` to be an array then 😬. But I will change to explicit undefined.

library/agent/ServiceConfig.ts
```diff
@@ -36,7 +36,18 @@ export class ServiceConfig {
     this.setAllowedIPAddresses(allowedIPAddresses);
   }
 
-  private setEndpoints(endpoints: Endpoint[]) {
+  private setEndpoints(endpointConfigs: EndpointConfig[]) {
+    const endpoints = endpointConfigs.map((endpoint) => {
+      return {
+        ...endpoint,
+        allowedIPAddresses:
```
bitterpanda63, line 43: would move to line 41 for readability? make it an if statement, just suggestion

bitterpanda63 (approved):

hansott (approved):

PR #565, opened 2025-03-28
Cache forceProtectionOff in context

timokoessler:


library/vulnerabilities/ssrf/inspectDNSLookupCalls.test.ts
```diff
@@ -86,7 +88,7 @@ t.test("it blocks lookup in blocking mode", (t) => {
     "operation"
   );
 
-  runWithContext(context, () => {
+  runWithContext(getContext(), () => {
```
bitterpanda63, line 91: is this related?

bitterpanda63: pff okay it's fine

timokoessler: Yes, otherwise the tests fail because we reuse the same context object and set forceProtectionOff.

library/agent/context/isProtectionOffForRoute.ts
```diff
@@ -0,0 +1,24 @@
+import type { Agent } from "../Agent";
+import { updateContext, type Context } from "../Context";
+
+export function isProtectionOffForRoute(
+  agent: Agent,
+  context: Readonly<Context> | undefined
+): boolean {
+  if (!context) {
+    return false;
```
bitterpanda63, line 9: don't we want to return true, if context is not set, we don't have userinput to compare to?

bitterpanda63 (commented):
🚀

library/agent/context/isProtectionOffForRoute.ts
timokoessler: The existing behavior was to still continue without context at the locations where this function is used.

bitterpanda63 (approved):

PR #570, opened 2025-04-03
Add Dev Container

timokoessler:
For an easier and safer development process

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,8 +20,7 @@
         "dbaeumer.vscode-eslint",
         "esbenp.prettier-vscode",
         "YoavBls.pretty-ts-errors",
-        "rust-lang.rust-analyzer",
-        "Shopify.ruby-lsp"
+        "rust-lang.rust-analyzer"
       ]
     }
   },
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,41 @@
...
+    // Configure properties specific to VS Code.
+    "vscode": {
+      "settings": {},
+      "extensions": [
+        "ms-azuretools.vscode-docker",
+        "dbaeumer.vscode-eslint",
+        "esbenp.prettier-vscode",
+        "YoavBls.pretty-ts-errors",
+        "rust-lang.rust-analyzer",
+        "Shopify.ruby-lsp"
```
hansott, line 24: not sure if we need these?

timokoessler: Only for the new hook system, but if we switch between branches, it's not nice to have different dev container configs resulting in rebuilding.

hansott (approved):

.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,36 @@
...
+	sudo apt-get install k6 -y
+fi
+
+# Install wrk, sqlite3
+sudo apt install wrk sqlite3 -y
+
+# Install npm packages, build and run containers
+npm i
+npm run build
+npm run containers
```
hansott, line 36:
```suggestion
npm run containers

```

PR #574, opened 2025-04-04
Only count requests for discovered routes

timokoessler:
Or in case of an attack

### DIFF
--- a/library/sources/http-server/createRequestListener.ts
+++ b/library/sources/http-server/createRequestListener.ts
@@ -105,11 +105,5 @@ function createOnFinishRequestHandler(
       // Only count the request if the route is discovered
       agent.getInspectionStatistics().onRequest();
     }
-
-    if (context && context.attackDetected) {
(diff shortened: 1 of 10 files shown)


library/sources/http-server/createRequestListener.ts
```diff
@@ -102,9 +102,13 @@ function createOnFinishRequestHandler(
       })
     ) {
       agent.onRouteExecute(context);
+      // Only count the request if the route is discovered
+      agent.getInspectionStatistics().onRequest();
+    } else if (context && context.attackDetected) {
+      // Also count the request if an attack was detected, also if the route is not discovered
+      agent.getInspectionStatistics().onRequest();
```
hansott, line 109: I prefer a single call to `onRequest(...)`, wdyt?
- easier to reason about
- if we move things around, it will less likely break

(I know the chance is small that we would call this twice because of the `else if`)

hansott (dismissed):

hansott (approved):

PR #577, opened 2025-04-11
Support fast-xml-parser v5

timokoessler:


library/sources/FastXmlParser.ts
```diff
@@ -38,18 +38,39 @@ export class FastXmlParser implements Wrapper {
...
           });
-        });
-      });
+        }
+      );
+
+      return {
+        ...exports,
+        XMLParser: wrappedParser,
+      };
```
hansott, line 73: would the approach for v5 also work for v4?

timokoessler: Why didn't I think of that? 😄

hansott (approved):

PR #583, opened 2025-04-17
Fix path traversal ignored if in pathname

timokoessler:


library/sources/HTTPServer.test.ts
```diff
@@ -719,3 +722,61 @@ t.test(
...
+  await new Promise<void>((resolve) => {
+    server.listen(3327, async () => {
+      const response = await new Promise((resolve, reject) => {
+        const req = http.request(
+          {
+            hostname: "localhost",
+            port: 3327,
+            path: "/../package.json", // Path traversal attempt
+            method: "GET",
+          },
```
hansott, line 753: we can't do this with fetch helper?

library/sources/HTTPServer.test.ts
```diff
@@ -78,6 +80,7 @@ t.beforeEach(() => {
 
 const http = require("http") as typeof import("http");
 const https = require("https") as typeof import("https");
+const { readFileSync } = require("fs");
```
hansott, line 83: typeof import, not needed?

library/sources/HTTPServer.test.ts
```diff
@@ -719,3 +722,61 @@ t.test(
     });
   }
 );
+
+t.test("it blocks path traversal in path", async (t) => {
+  const server = http.createServer((req, res) => {
+    try {
+      // @ts-expect-error Ignore
+      const file = readFileSync(join(__dirname, req.url));
```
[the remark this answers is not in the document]
timokoessler: `req.pathname` does not exist. If I parse the URL with new URL, it will be normalized again.

PR #588, opened 2025-04-21
Allow passing a Router to `addExpressMiddleware`

### LINKED ISSUE #587: Cannot apply the middleware on a Router
The Express app can be composed of one or more `express.Router` instances as described [here](https://expressjs.com/en/guide/routing.html) (bottom of the page). This is a best practice as it promotes increased modularity.

However, the current approach of applying the Zen middleware does not seem to support this use case, as `addExpressMiddleware` does not accept the router as input. 

Image

Is there a specific reason for having Zen apply the middleware instead of the more common practice of exporting the middleware that can be applied using `app.use(zen)`/`router.use(zen)`? Examples: [the express guide](https://expressjs.com/en/guide/using-middleware.html), [express-rate-limit](https://github.com/express-rate-limit/express-rate-limit/blob/main/source/lib.ts#L324), [auth0/express-jwt](https://github.com/auth0/express-jwt/blob/master/src/index.ts#L106)

hansott:
Also log warning when middleware is called multiple times during a request.

Closes #587

### DIFF
--- a/library/middleware/shouldBlockRequest.ts
+++ b/library/middleware/shouldBlockRequest.ts
@@ -23,7 +23,6 @@ export function shouldBlockRequest(): Result {
 
   if (context.executedMiddleware) {
     logWarningAlreadyExecutedMiddleware();
-    return { block: false };
   }
 
   updateContext(context, "executedMiddleware", true);
--- a/library/middleware/shouldBlockRequest.ts
+++ b/library/middleware/shouldBlockRequest.ts
@@ -21,6 +21,11 @@ export function shouldBlockRequest(): Result {
     return { block: false };
   }
 
+  if (context.executedMiddleware) {
+    logWarningAlreadyExecutedMiddleware();
+    return { block: false };
+  }
+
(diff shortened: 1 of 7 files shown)


library/middleware/shouldBlockRequest.ts
```diff
@@ -21,6 +21,11 @@ export function shouldBlockRequest(): Result {
     return { block: false };
   }
 
+  if (context.executedMiddleware) {
+    logWarningAlreadyExecutedMiddleware();
+    return { block: false };
+  }
```
hansott, line 27: Should be fine?

willem-delbare (approved):

library/middleware/shouldBlockRequest.ts
timokoessler: Should we add an early return? Would be nice if we could return early and remove the logged warning, but what if the user adds the middleware twice and the `Zen.setUser` call is in between of both? In this case we would never do user blocking.

timokoessler (approved):

PR #590, opened 2025-04-28
Add Bot Spoofing Protection

timokoessler:


library/sources/http-server/createRequestListener.ts
```diff
@@ -53,7 +53,7 @@ function callListenerWithContext(
 ) {
   const context = contextFromRequest(req, body, module);
 
-  return runWithContext(context, () => {
+  return runWithContext(context, async () => {
```
hansott, line 56: could this change behaviour of errors being thrown inside?

library/vulnerabilities/bot-spoofing/verifyBotAuthenticityWithDNS.ts
```diff
@@ -0,0 +1,87 @@
...
+ *   e.g. `54.236.1.12` → crawl-54-236-1-12.pinterest.com
+ * 2. Check if the domain matches the expected origin
+ *   e.g. Domain is pinterest.com or pinterestcrawler.com for `Pinterestbot`
+ * 3. Because PTR records can be spoofed, also lookup the A and AAAA record and compare them with the request ip:
+ *   e.g. `crawl-54-236-1-12.pinterest.com` → `54.236.1.12`
+ */
+export async function verifyBotAuthenticityWithDNS(
+  requestIp: string,
+  matchingBot: ServiceConfigBotSpoofingData
+) {
```
hansott, line 29: Return early if request IP empty?

timokoessler: Already done in `checkContextForBotSpoofing`

library/sources/http-server/createRequestListener.ts
timokoessler: If an exception is thrown in the callback passed to `createServer`, the Node.js process crashes. The behavior is the same with and without `async`. It needs to be caught using try / catch inside the handler function itself.

library/vulnerabilities/bot-spoofing/verifyBotAuthenticityWithDNS.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (
+      error instanceof Error &&
+      (error as NodeJS.ErrnoException).code === "ENOTFOUND"
+    ) {
+      // No matching hostnames found, so the bot is not authentic
+      cache.set(cacheKey, false);
+      return false;
+    }
+
+    getInstance()?.log(`Bot Spoofing Protection: DNS check failed: ${error}`);
```
hansott, line 84: Do we want to log something in the console without debug enabled?

hansott: already good!

library/sources/http-server/checkIfRequestIsBlocked.ts
```diff
@@ -147,5 +148,14 @@ export function checkIfRequestIsBlocked(
     return true;
   }
 
+  const isBotSpoofing = await checkContextForBotSpoofing(context, agent);
+  if (isBotSpoofing) {
+    res.statusCode = 403;
+    res.setHeader("Content-Type", "text/plain");
+
+    res.end("You are not allowed to access this resource.");
```
hansott, line 158: Should we add a reason, prob not right?

hansott (approved):

library/sources/http-server/checkIfRequestIsBlocked.ts
timokoessler: Would allow easier debugging, but also gives an attacker more info.

hansott: Yup, indeed! Let's keep it like this

PR #595, opened 2025-05-02
Fix absolute path traversal bypass with backslash conversion

hansott:
I also found out we didn't check \t \r \n for absolute paths (if it's an URL object)

### DIFF
--- a/library/vulnerabilities/path-traversal/detectPathTraversal.ts
+++ b/library/vulnerabilities/path-traversal/detectPathTraversal.ts
@@ -2,7 +2,10 @@ import {
   containsUnsafePathParts,
   containsUnsafePathPartsUrl,
 } from "./containsUnsafePathParts";
-import { startsWithUnsafePath } from "./unsafePathStart";
+import {
+  startsWithUnsafePath,
+  startsWithUnsafePathUrl,
+} from "./unsafePathStart";
 import { fileURLToPath } from "url";
 
 export function detectPathTraversal(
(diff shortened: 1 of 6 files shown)


library/vulnerabilities/path-traversal/normalizeLikeURLConstructor.ts
```diff
@@ -0,0 +1,14 @@
...
+ * The WHATWG URL spec defines the following:
+ * - Remove all ASCII tab or newline from input.
+ * - An ASCII tab or newline is U+0009 TAB, U+000A LF, or U+000D CR.
+ *
+ * Also, backward slashes are converted to forward slashes by the URL constructor.
+ *
+ * See https://url.spec.whatwg.org/#url-parsing
+ */
+export function normalizeLikeURLConstructor(url: string): string {
+  return url.replace(/[\t\n\r]/g, "").replace(/\\/g, "/");
```
timokoessler, line 13: Performance wise one regex might be quicker, but this is more readable.

library/vulnerabilities/path-traversal/detectPathTraversal.ts
```diff
@@ -19,13 +22,18 @@ export function detectPathTraversal(
...
   // Also /./ is checked by normal absolute path traversal check (if #219 is merged)
   // Use containsUnsafePathPartsUrl, because urls can contain a TAB, carriage return or line feed that is silently removed by the URL constructor.
-  if (isUrl && containsUnsafePathPartsUrl(userInput)) {
-    const filePathFromUrl = parseAsFileUrl(userInput);
-    if (filePathFromUrl && filePath.includes(filePathFromUrl)) {
-      return true;
+  // Use startsWithUnsafePathUrl, because URLs can contain backward slashes that are converted to forward slashes by the URL constructor.
+  if (isUrl) {
+    const containsUnsafePath = containsUnsafePathPartsUrl(userInput);
+    const startWithUnsafePath = startsWithUnsafePathUrl(filePath, userInput);
```
timokoessler, line 30: Only check if `checkPathStart` is true?

library/vulnerabilities/path-traversal/normalizeLikeURLConstructor.ts
hansott:
```js
function normalizeLikeURLConstructor1(url) {
  return url.replace(/[\t\n\r]/g, "").replace(/\\/g, "/");
}

function normalizeLikeURLConstructor2(url) {
  return url.replace(/[\t\n\r\\]/g, (match) => (match === "\\" ? "/" : ""));
}

const samples = ["/etc/passwd", "\\etc/passwd", "/etc\t/passwd"];

const start1 = performance.now();
for (let i = 0; i < 1000000; i++) {
  for (const sample of samples) {
    normalizeLikeURLConstructor1(sample);
  }
}
const end1 = performance.now();
console.log(
  "normalizeLikeURLConstructor1 Average time:",
  (end1 - start1) / 1000000,
  "ms"
);

const start2 = performance.now();
for (let i = 0; i < 1000000; i++) {
  for (const sample of samples) {
    normalizeLikeURLConstructor2(sample);
  }
}
const end2 = performance.now();
console.log(
  "normalizeLikeURLConstructor2 Average time:",
  (end2 - start2) / 1000000,
  "ms"
);
```

```
normalizeLikeURLConstructor1 Average time: 0.00020970724999999998 ms
normalizeLikeURLConstructor2 Average time: 0.000271266542 ms
```

library/vulnerabilities/path-traversal/detectPathTraversal.ts
hansott: Thanks, fixed!

timokoessler (approved):

PR #599, opened 2025-05-12
Improve reliability of path traversal detection

timokoessler:


library/vulnerabilities/path-traversal/checkUrlPathForPathTraversal.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (!url.includes("%")) {
+    // Performance optimization, if the URL does not contain any encoded characters, we don't need to decode it
+    return {
+      found: false,
+    };
+  }
+
+  // Also check encoded paths
+  const decodedPath = decodeURIComponent(rawPath);
```
timokoessler, line 45: Does not cover double-encoding, should we?

bitterpanda63: Question for @kapyteinaikido

library/vulnerabilities/path-traversal/checkUrlPathForPathTraversal.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  if (!url.includes("%")) {
+    // Performance optimization, if the URL does not contain any encoded characters, we don't need to decode it
+    return {
+      found: false,
+    };
+  }
+
+  // Also check encoded paths
+  const decodedPath = normalizeLikeURLConstructor(decodeURIComponent(rawPath));
```
hansott, line 46: Decode can throw?

library/vulnerabilities/path-traversal/checkUrlPathForPathTraversal.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  if (forbiddenPattern.test(rawPath)) {
+    return {
+      found: true,
+      payload: rawPath,
+    };
+  }
+
+  if (!url.includes("%")) {
```
hansott, line 38: We could also add a length check? Since we expect at least two chars behind `%`?

hansott: Nevermind already checked above 👍

library/vulnerabilities/path-traversal/checkUrlPathForPathTraversal.ts
timokoessler: Fixed

PR #628, opened 2025-06-06
Prevent agent from starting twice in edge case

timokoessler:
If the Node.js require cache is cleared, it is possible to execute a js script a second time, while the old one also keeps running.


library/helpers/indexImportGuard.ts
```diff
@@ -0,0 +1,19 @@
+const symbol = Symbol.for("zen.nodejs.index.run.once");
+
+export function checkIndexImportGuard(): boolean {
+  const globalState = globalThis as typeof globalThis & {
```
hansott, line 4: does this work in ESM? but prob not relevant for ESM? Since you cannot clear cache I guess?

timokoessler: Also works with ESM, as global object is the same in CJS and ESM. Not sure what this would do tho: https://nodejs.org/api/module.html#moduleflushcompilecache

library/helpers/indexImportGuard.test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  // Clear require cache
+  for (const path in require.cache) {
+    delete require.cache[path];
+  }
+
+  require("../index");
+
+  t.same(logs, [
+    "AIKIDO: Zen has already been initialized. Please ensure that Zen is imported only once in your application, as importing it multiple times can lead to unexpected behavior.",
```
hansott: You can import Zen multiple times? e.g. for using `setUser` etc

Maybe tweak the message a bit? To focus on the part that require might have been cleared or something else weird is going on?

PR #629, opened 2025-06-08
Add support for OpenAI SDK

hansott:


library/agent/api/Event.ts
```diff
@@ -112,6 +113,26 @@ type Heartbeat = {
     ipAddresses: {
       breakdown: Record<string, number>;
     };
+    ai: {
```
timokoessler: Basically the `AIProviderStats` type without `routesLRU`? Maybe nicer to not define inline, so it's e.g. easier visible that it's an array?

library/sinks/OpenAI.ts
```diff
@@ -0,0 +1,129 @@
...
+    const context = getContext();
+
+    if (context && context.route && context.method) {
+      return { path: context.route, method: context.method };
+    }
+
+    return undefined;
+  }
+
+  private inspectCompletionsResponse(agent: Agent, completion: unknown) {
```
timokoessler: Method never called?

library/agent/api/Event.ts
hansott: Removed routes altogether

library/sinks/OpenAI.ts
hansott: Removed it for now, I was still working on the wrapping

willem-delbare (approved):

library/sinks/AwsSDKVersion3.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export class AwsSDKVersion3 implements Wrapper {
+  private processResponse(response: unknown, agent: Agent) {
+    if (!isResponse(response)) {
+      return;
+    }
+
+    let body;
+    try {
+      const json = new TextDecoder().decode(response.body);
```
timokoessler: Not sure if we should set `fatal` to true, like also done in [library/vulnerabilities/path-traversal/checkContextForPathTraversal.ts](https://github.com/AikidoSec/firewall-node/blob/e4986e39bf7d9e140637c3f7287236d2bb67f9bb/library/vulnerabilities/path-traversal/checkContextForPathTraversal.ts#L66)? See https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/fatal

library/sinks/AwsSDKVersion3.ts
```diff
@@ -0,0 +1,89 @@
...
+          wrapExport(exports.BedrockRuntimeClient.prototype, "send", pkgInfo, {
+            kind: "ai_op",
+            modifyReturnValue: (args, returnValue, agent) => {
+              if (args.length > 0) {
+                if (
+                  returnValue instanceof Promise &&
+                  args[0] instanceof exports.InvokeModelCommand
+                ) {
+                  returnValue.then((response) => {
+                    this.processResponse(response, agent);
```
timokoessler: Because logic is called inside .then, exceptions might not be caught by try ... catch in the agents wrapping logic?

README.md
```diff
@@ -98,6 +98,10 @@ See list above for supported database drivers.
 
 * ✅ [`@koa/router`](https://www.npmjs.com/package/@koa/router) 13.x, 12.x, 11.x and 10.x
 
+### AI SDKs
+
```
timokoessler, line 102: Only monitoring, no attack protection, e.g. unsafe tool usage?

PR #635, opened 2025-06-12
Do not wrap exec as it internally uses execFile

timokoessler:


library/sinks/ChildProcess.ts
```diff
@@ -44,6 +38,7 @@ export class ChildProcess implements Wrapper {
         },
       });
       wrapExport(exports, "execFile", pkgInfo, {
+        // Also protects exec
```
hansott:
```suggestion
        // Also protects exec, see https://github.com/nodejs/node/blob/main/lib/child_process.js
```

PR #636, opened 2025-06-13
Add rate limiting groups

timokoessler:


library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -57,13 +57,14 @@ export function shouldRateLimitRequest(
...
       .getRateLimiter()
-      .isAllowed(
-        `${endpoint.method}:${endpoint.route}:user:${context.user.id}`,
-        windowSizeInMS,
-        maxRequests
-      );
+      .isAllowed(key, windowSizeInMS, maxRequests);
 
     if (!allowed) {
       return { block: true, trigger: "user" };
```
timokoessler: Not sure if we should change trigger? Would also be a change for users using a custom middleware.

library/agent/Users.ts
```diff
@@ -48,6 +52,7 @@ export class Users {
       return {
         id: user.id,
         name: user.name,
+        rateLimitGroup: user.rateLimitGroup,
```
hansott: do we want to send this in a heartbeat?

library/ratelimiting/shouldRateLimitRequest.ts
hansott: It would be nice that the `trigger` would be `group` but indeed, it might break custom middlewares... given the security context, safer to keep it like this?

library/agent/Users.ts
timokoessler: Removed

library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -56,6 +60,23 @@ export function shouldRateLimitRequest(
...
+        `${endpoint.method}:${endpoint.route}:group:${context.rateLimitGroup}`,
+        windowSizeInMS,
+        maxRequests
+      );
+
+    if (!allowed) {
+      return { block: true, trigger: "group" };
+    }
+
+    // Do not check IP rate limit if user is set
```
bitterpanda63: ?

library/ratelimiting/shouldRateLimitRequest.ts
```diff
@@ -56,6 +60,23 @@ export function shouldRateLimitRequest(
...
+        `${endpoint.method}:${endpoint.route}:group:${context.rateLimitGroup}`,
+        windowSizeInMS,
+        maxRequests
+      );
+
+    if (!allowed) {
+      return { block: true, trigger: "group" };
+    }
+
+    // Do not check IP rate limit if rateLimitGroup is set
```
bitterpanda63:
```suggestion
    // Do not check IP or User rate limit if rateLimitGroup is set
```

library/ratelimiting/group.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (!context) {
+    logWarningSetRateLimitGroupCalledWithoutContext();
+    return;
+  }
+
+  const group = g as unknown;
+
+  if (!isPlainObject(group)) {
+    agent.log(
+      `setRateLimitGroup(...) expects an object with 'id' properties, found ${typeof group} instead.`
```
hansott:
```suggestion
      `setRateLimitGroup(...) expects an object with 'id' property, found ${typeof group} instead.`
```

library/sources/Hono.test.ts
```diff
@@ -99,6 +110,10 @@ function getApp() {
       setUser({ id: "567" });
     } else if (c.req.path.startsWith("/user")) {
       setUser({ id: "123" });
+    } else if (c.req.path.startsWith("/rate-limited-group")) {
+      const userId = c.req.header("X-User-Id") || "123";
```
hansott: Unused?

hansott: Need to set user ID? would be nice to test combination

hansott (commented):
🚀

library/sources/Hono.test.ts
timokoessler: Fixed, unit tests have tests with combinations?

hansott (approved):

PR #638, opened 2025-06-16
Add rate limiting stats

timokoessler:


library/sources/http-server/http2/createStreamListener.ts
```diff
@@ -56,17 +56,19 @@ function discoverRouteFromStream(
...
-      // Only count the request if the route is discovered
-      agent.getInspectionStatistics().onRequest();
+    if (!isNaN(statusCode)) {
+      if (
+        shouldDiscoverRoute({
+          statusCode: statusCode,
+          route: context.route,
+          method: context.method,
+        }) ||
+        context.rateLimited // Also discover routes for rate-limited requests
```
hansott: hmm, when does this happen? I think we don't need it:

should discover route -> the route will show up in the dashboard
user added route manually and then added rate limiting -> if the route is a wildcard, it would not include the rate limited count on the wildcard route? e.g. `ANY /api/*` then the route would be `POST /api/login`, so `POST /api/login` would contain the rate limit counts?

library/sources/http-server/createRequestListener.ts
```diff
@@ -91,19 +91,19 @@ function createOnFinishRequestHandler(
...
-      // Only count the request if the route is discovered
-      agent.getInspectionStatistics().onRequest();
+    if (context && context.route && context.method) {
+      if (
+        shouldDiscoverRoute({
+          statusCode: res.statusCode,
+          route: context.route,
+          method: context.method,
+        }) ||
+        context.rateLimited // Also discover routes for rate-limited requests
```
hansott: [see other comment](https://github.com/AikidoSec/firewall-node/pull/638/files#r2190321416)

hansott (approved):

PR #643, opened 2025-06-17
Re-add routes to AI stats

timokoessler:


library/sinks/OpenAI.ts
```diff
@@ -113,6 +116,16 @@ export class OpenAI implements Wrapper {
     return "openai";
   }
 
+  private getRoute() {
```
bitterpanda63: maybe extract as a helper function? we will probably expand AI usage coverage right

library/sinks/AwsSDKVersion3.ts
```diff
@@ -113,9 +115,20 @@ export class AwsSDKVersion3 implements Wrapper {
       model: modelId,
       inputTokens: inputTokens,
       outputTokens: outputTokens,
+      route: this.getRoute(),
     });
   }
 
+  private getRoute() {
```
bitterpanda63: same here

library/agent/api/Event.ts
```diff
@@ -123,6 +123,16 @@ type Heartbeat = {
       output: number;
       total: number;
     };
+    routes: {
```
bitterpanda63, line 126: we could also add ai usage stats to the Routes object, like we (are planning to) do for the rate limiting stats?

library/agent/AIStatistics.ts
```diff
@@ -36,20 +55,54 @@ export class AIStatistics {
...
           total: 0,
         },
+        routesLRU: new LRUMap<string, AIRouteStats>(this.maxRoutes),
       });
     }
 
     return this.calls.get(key)!;
   }
 
+  private ensureRouteStats(
```
bitterpanda63, line 65: If we use the existing Routes object would also remove need for this (and the LRU)

library/sinks/OpenAI.ts
timokoessler: Yes, right

library/sinks/AwsSDKVersion3.ts
```diff
@@ -2,6 +2,7 @@ import { Agent } from "../agent/Agent";
 import { Hooks } from "../agent/hooks/Hooks";
 import { wrapExport } from "../agent/hooks/wrapExport";
 import { Wrapper } from "../agent/Wrapper";
+import { getRouteForAiStats } from "../helpers/getRouteForAIStats";
```
bitterpanda63:
```suggestion
import { getRouteForAiStats } from "../helpers/getRouteForAiStats";
```

library/sinks/OpenAI.ts
```diff
@@ -3,6 +3,7 @@ import { Hooks } from "../agent/hooks/Hooks";
 import { Wrapper } from "../agent/Wrapper";
 import { wrapExport } from "../agent/hooks/wrapExport";
 import { isPlainObject } from "../helpers/isPlainObject";
+import { getRouteForAiStats } from "../helpers/getRouteForAIStats";
```
bitterpanda63:
```suggestion
import { getRouteForAiStats } from "../helpers/getRouteForAiStats";
```

bitterpanda63 (approved):

PR #645, opened 2025-06-26
Support OpenAI v5

timokoessler:


library/sinks/OpenAI.ts
```diff
@@ -113,14 +113,37 @@ export class OpenAI implements Wrapper {
     return "openai";
   }
 
+  private getResponsesClass(exports: any) {
```
hansott, line 116: Use `unknown` rather than `any`?

library/sinks/OpenAI.ts
```diff
@@ -113,14 +113,37 @@ export class OpenAI implements Wrapper {
...
+    if (exports.Responses) {
+      return exports.Responses;
+    }
+    if (exports.OpenAI && exports.OpenAI.Responses) {
+      return exports.OpenAI.Responses;
+    }
+  }
+
+  private getCompletionsClass(exports: any) {
```
hansott, line 125: Use `unknown` rather than `any`?

library/sinks/OpenAI.ts
```diff
@@ -113,14 +113,37 @@ export class OpenAI implements Wrapper {
...
+    if (exports.Chat && exports.Chat.Completions) {
+      return exports.Chat.Completions;
+    }
+    if (
+      exports.OpenAI &&
+      exports.OpenAI.Chat &&
+      exports.OpenAI.Chat.Completions
+    ) {
+      return exports.OpenAI.Chat.Completions;
+    }
```
hansott, line 135: Add comments if the `if` statement is about v4 or v5?

PR #646, opened 2025-06-26
Block outbound HTTP(s) requests

timokoessler:


end2end/tests-new/hono-pg-esm-outbound.test.mjs
bitterpanda63, line 1: why is that directory called tests-new btw?

library/agent/ServiceConfig.ts
```diff
@@ -278,4 +281,24 @@ export class ServiceConfig {
...
+
+  updateDomains(domains: Domain[]) {
+    this.domains = new Map(domains.map((i) => [i.hostname, i.mode]));
+  }
+
+  shouldBlockOutgoingRequest(hostname: string): boolean {
+    const mode = this.domains.get(hostname);
+
+    if (this.blockNewOutgoingRequests) {
+      // Only allow outgoing requests if the mode is "allow"
```
bitterpanda63, line 297: I don't understand this part exactly if the mode is "block" it will always return true regardless of blockNewOutgoingRequests, if the mode is "allow" it will always allow? why do we do the check then?

end2end/tests-new/hono-pg-esm-outbound.test.mjs
timokoessler: We use two separate test runners for the old and new e2e tests :). The reason was a bug preventing ESM tests using the old one and that we would like to switch.

library/agent/ServiceConfig.ts
hansott: As discussed in Slack, I'll improve this!

hansott: Added another comment to make the intent more clear: https://github.com/AikidoSec/firewall-node/pull/646/commits/e17876f094909353d4bc30ebe700dfed97e7f29b

I tried to rewrite this but ended up with a bunch of more code that wasn't better

PR #648, opened 2025-07-02
Add support for restify framework ^8.0.0

hansott:


library/middleware/restify.ts
```diff
@@ -0,0 +1,38 @@
...
+        let message = "You are rate limited by Zen.";
+        if (result.trigger === "ip" && result.ip) {
+          message += ` (Your IP: ${escapeHTML(result.ip)})`;
+        }
+
+        res.status(429);
+        res.setHeader("Content-Type", "text/plain");
+        res.send(message);
+
+        return next(false);
```
timokoessler, line 24: It does not call the next handlers?

library/sources/Restify.v8.test.ts
```diff
@@ -0,0 +1,16 @@
+import * as t from "tap";
+import { getMajorNodeVersion } from "../helpers/getNodeVersion";
+import { createRestifyTests } from "./Restify.tests";
+
+t.test(
+  "Koa",
```
timokoessler: Koa?

library/middleware/restify.ts
hansott: Yeah, we need to cancel the rest of the request 😊

library/sources/Restify.v8.test.ts
hansott: Oohhh, thanks! Fixed!

PR #663, opened 2025-07-11
Add AI timing for OpenAI

bitterpanda63:


hansott:
keep the branch around, let's revisit later

PR #664, opened 2025-07-14
Ignore alphanumerical payloads ending with =

hansott:
query
```sql
SELECT * FROM table WHERE table.active= 1
```

payload
```e=```

will alter the token structure after replacing with safe string "aa":
```sql
SELECT * FROM table WHERE table.activaa 1
```

So it will be flagged as a SQL injection, in general just using `e=` as payload will be very difficult to exploit. (as the generated SQL query should still be valid)

Waiting for https://github.com/AikidoSec/zen-internals/pull/90


hansott:
See https://github.com/AikidoSec/zen-internals/pull/90

PR #670, opened 2025-07-21
Fix file url path traversal bypass with control chars

timokoessler:


library/vulnerabilities/path-traversal/detectPathTraversal.ts
```diff
@@ -75,3 +75,18 @@ function parseAsFileUrl(path: string) {
   }
   return undefined;
 }
+
+/**
+ * Checks if a string starts with "file:" to determine if it is a file URL.
+ * Removes control characters at the start of the path and checks lowercase.
```
hansott:
```suggestion
 * Removes control characters at the start of the path and checks lowercase.
 * See https://url.spec.whatwg.org/#url-parsing
```

library/vulnerabilities/path-traversal/detectPathTraversal.ts
```diff
@@ -75,3 +75,18 @@ function parseAsFileUrl(path: string) {
...
+ */
+function isFileUrlString(path: string): boolean {
+  // Remove control characters at the start of the path
+  return (
+    path
+      // oxlint-disable-next-line no-control-regex
+      .replace(/^[\x00-\x1F\x7F\x20]+/, "")
```
hansott: > A C0 control is a [code point](https://infra.spec.whatwg.org/#code-point) in the range U+0000 NULL to U+001F INFORMATION SEPARATOR ONE, inclusive.
> A C0 control or space is a [C0 control](https://infra.spec.whatwg.org/#c0-control) or U+0020 SPACE.

I don't think we should include `\x7F`?

```js
> new URL('\u007Ffile://path')
Uncaught TypeError: Invalid URL
    at new URL (node:internal/url:787:36) {
  code: 'ERR_INVALID_URL',
  input: '\x7Ffile://path'
}
```

Additionally, I would prefer this notation I think (more like the spec notation), although `\x` notation works mostly the same:

```suggestion
      .replace(/^[\u0000-\u0020]+/, "")
```

Can still be split up (to show that space is included):

```suggestion
      .replace(/^[\u0000-\u001F\u0020]+/, "")
```

PR #675, opened 2025-07-24
Send heartbeat on shutdown

timokoessler:
## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Added shutdown wrapper file that invoked AgentSingleton.getInstance().shutdown with optional timeout
* Exported shutdown from library index.ts in both named and default exports

<sup>[More info](https://app.aikido.dev/featurebranch/scan/109507951?groupId=6)</sup>


library/agent/Agent.ts
```diff
@@ -609,4 +612,27 @@ export class Agent {
...
+        }
+        this.shutdownEventCalled = true;
+        this.onShutdown();
+      });
+    });
+  }
+
+  private async onShutdown() {
+    this.logger.log("Shutting down agent...");
+    if (performance.now() - this.lastHeartbeat > 30000) {
```
hansott: then we'll still miss some data? would make sure that a server keeps crashing and restarting, we don't send a heartbeat every time, which seems to be the case now? so rather use the start date as reference?

library/agent/Agent.ts
```diff
@@ -609,4 +612,27 @@ export class Agent {
         return this.sendHeartbeatEveryMS;
     }
   }
+
+  private registerShutdownHandler() {
+    const shutdownEvents = ["SIGTERM", "SIGINT", "SIGBREAK", "SIGHUP"];
+
+    shutdownEvents.forEach((event) => {
+      process.on(event, () => {
```
hansott: Does this result in node.js not exiting like default?

timokoessler: Yes, that is why process.exit is called

library/agent/Agent.ts
timokoessler: Changed

docs/graceful-shutdown.md
```diff
@@ -0,0 +1,20 @@
+# Graceful Shutdown
+
+If you already registered a shutdown handler, e.g. using `process.on('SIGTERM', ...)`, you can call `await Zen.shutdown()` to ensure the latest stats are sent before the process exits.
+See the example below for how to implement this, if you don't already have a shutdown handler in place.
+Please note that this might not work correctly in all environments, especially on Windows.
```
hansott:
```suggestion
Please note that this might not work correctly on Windows, due to [differences in how signals like SIGTERM and SIGINT are handled](https://nodejs.org/api/process.html#signal-events).
```

docs/graceful-shutdown.md
```diff
@@ -0,0 +1,20 @@
...
+  // Perform any cleanup or final tasks here
+  await Zen.shutdown();
+
+  process.exit(process.exitCode || 0);
+}
+
+process.on("SIGTERM", shutdownHandler);
+process.on("SIGINT", shutdownHandler);
+// Handle other signals as needed
+```
```
hansott, line 19: People will not do this by default... should we mention this in the framework docs and link to here?

library/agent/shutdown.ts
```diff
@@ -0,0 +1,5 @@
+import { getInstance } from "./AgentSingleton";
+
+export async function shutdown() {
```
hansott:
```suggestion
export async function shutdown(): Promise<void> {
```

hansott: mehh, it's inferred by TS but thought it was a good idea to enforce that this function doesn't return a value?

library/agent/Agent.ts
```diff
@@ -609,4 +610,12 @@ export class Agent {
         return this.sendHeartbeatEveryMS;
     }
   }
+
+  public async shutdown() {
+    this.logger.log("Shutting down agent...");
+    if (performance.now() > 30000) {
+      // Only send heartbeat if we are running for more than 30 seconds
+      await this.flushStats(1000);
+    }
```
hansott, line 724: maybe one extra safety check that we didn't just send a heartbeat? e.g. at least one second?

timokoessler:
LGTM

hansott (approved):

PR #677, opened 2025-07-28
Set original function in `__NR_original` property

hansott:
This ensures that loopback framework can find it back:

```js
/*!
 * Find the corresponding express layer by handler *
 * This is needed because monitoring agents such as NewRelic can add handlers
 * to the stack. For example, NewRelic adds sentinel handler. We need to search
 * the stackto find the correct layer. */ proto._findLayerByHandler = function(handler) {
  // Other handlers can be added to the stack, for example,
  // NewRelic adds sentinel handler, and AppDynamics adds
  // some additional proxy info. We need to search the stack
  for (let k = this._router.stack.length - 1; k >= 0; k--) {
    const isOriginal = this._router.stack[k].handle === handler;
    const isNewRelic = this._router.stack[k].handle['__NR_original'] === handler;
    const isAppDynamics = this._router.stack[k].handle['__appdynamicsProxyInfo__'] &&
      this._router.stack[k].handle['__appdynamicsProxyInfo__']['orig'] === handler;
```

`this._router.stack[k].handle === handler` doesn't work since we wrap the functions and that changes the identity.

### DIFF
--- a/library/helpers/wrap.ts
+++ b/library/helpers/wrap.ts
@@ -32,6 +32,7 @@ export function createWrappedFunction(
   const wrapped = wrapper(original);
 
   defineProperty(wrapped, "__original", original);
+  defineProperty(wrapped, "__NR_original", original);
   defineProperty(wrapped, "__wrapped", true);
 
   // Copy over all properties from the original function to the wrapped one.


library/helpers/wrap.ts
```diff
@@ -32,6 +32,7 @@ export function createWrappedFunction(
   const wrapped = wrapper(original);
 
   defineProperty(wrapped, "__original", original);
+  defineProperty(wrapped, "__NR_original", original);
```
timokoessler, line 35: In case the patch solution is rejected, let's do this not by default, only e.g. if some env var is set or we detect that framework?

hansott:
Closing this, we can provide a patch for loopback framework.

PR #679, opened 2025-07-31
Add oxlint

timokoessler:
In parallel with eslint for now


library/helpers/ip-matcher/Network.ts
```diff
@@ -121,6 +121,7 @@ export class Network {
     if (this.addr.bytes().length !== network.addr.bytes().length) return false;
 
     // handle edge cases
+    // oxlint-disable-next-line eqeqeq
```
hansott: Fix `network.netbits === 0`?

timokoessler: Done

PR #684, opened 2025-08-04
Export fastify hook that can be used in `preHandler`

hansott:
`onRequest` runs quite early in the fastify request lifecycle... some users rely on the `preHandler` phase for authentication. Let's export the fastify hook itself so that a user can use this on a route in `preHandler`.


timokoessler:
Sorry 😢

PR #685, opened 2025-08-04
Use --ignore-scripts for install script

hansott:


scripts/install.js
```diff
@@ -61,6 +64,69 @@ async function installDependencies(folder) {
...
+ */
+async function rebuildNativePackages(folder) {
+  const packageJsonPath = join(projectRoot, folder, "package.json");
+
+  if (!(await fileExists(packageJsonPath))) {
+    return;
+  }
+
+  const buffer = await readFile(packageJsonPath, "utf-8");
+  const pkg = JSON.parse(buffer.toString());
```
timokoessler: Buffer is already a string because of "utf-8"?

scripts/install.js
```diff
@@ -61,6 +64,69 @@ async function installDependencies(folder) {
...
+  const allDeps = {
+    ...(pkg.dependencies ?? {}),
+    ...(pkg.devDependencies ?? {}),
+  };
```
timokoessler: Sets are nice (no need to change)

hansott:
See #712

PR #686, opened 2025-08-05
Add macOS root dirs to path traversal check

timokoessler:


library/vulnerabilities/path-traversal/unsafePathStart.ts
```diff
@@ -1,28 +1,38 @@
...
+  "var",
+  // macOS specific
+  "applications",
+  "cores",
+  "library",
+  "private",
+  "users",
+  "system",
+  "volumes",
+].map((f) => `/${f}/`);
```
hansott: I think for search-ability, it's better to have the slashes present, also reduces the diff?

timokoessler: Changed

library/vulnerabilities/path-traversal/unsafePathStart.ts
```diff
@@ -21,8 +22,17 @@ const linuxRootFolders = [
...
   "/var/",
+  // macOS specific
+  "/applications/",
+  "/cores/",
+  "/library/",
+  "/private/",
+  "/users/",
+  "/system/",
+  "/volumes/",
 ];
```
hansott:
```suggestion
].map(path => path.toLowerCase());
```

Then we can remove the comment? :shipit:

PR #699, opened 2025-08-14
Handle IPv6 addresses with ports correctly

timokoessler:


library/helpers/getIPAddressFromRequest.ts
```diff
@@ -44,9 +44,15 @@ function getClientIpFromHeader(value: string) {
     // According to https://www.rfc-editor.org/rfc/rfc7239 (5.2) X-Forwarded-For
     // is allowed to include a port number, so we check this here :
     if (ip.includes(":")) {
-      const parts = ip.split(":");
+      // If the ip is an IPv6 in the format [ipv6]:port, we need to split it using ]:
+      const splitWith = ip.startsWith("[") ? "]:" : ":";
```
bitterpanda63: this notation is very hard to read with all the : and the [ and the ( haha

library/helpers/getIPAddressFromRequest.ts
```diff
@@ -44,13 +44,28 @@ function getClientIpFromHeader(value: string) {
...
+          return parts[0].slice(1);
+        }
         return parts[0];
       }
     }
 
+    // Normalize IPv6 by removing brackets
+    if (ip.startsWith("[") && ip.endsWith("]")) {
+      return ip.slice(1, -1);
+    }
```
hansott:
```js
const forwardedIps = value.split(",").map((e) => {
  const ip = e.trim();

  // We do a first check here to make sure that valid IPv6 addresses don't
  // get split on ":" below.
  if (isIP(ip)) {
    return ip;
  }

  // According to https://www.rfc-editor.org/rfc/rfc7239 (5.2) X-Forwarded-For
  // is allowed to include a port number

  if (ip.startsWith("[") && ip.endsWith("]")) {
    // IPv6 without port: [ip]
    return ip.slice(1, -1);
  }

  if (ip.startsWith("[")) {
    // IPv6 with port: [ip]:port
    const closingBracket = ip.indexOf("]:");
    if (closingBracket > 0) {
      return ip.slice(1, closingBracket);
    }
  }

  // Handle IPv4 with port: ip:port
  if (ip.includes(":")) {
    const parts = ip.split(":");
    if (parts.length === 2) {
      return parts[0];
    }
  }

  return ip;
});
```

Perhaps we can write it out so that the different cases become very obvious?

hansott (approved):

library/helpers/getIPAddressFromRequest.ts
hansott: See https://github.com/AikidoSec/firewall-node/pull/699/commits/3eb27a1032846cff2144e69b9e8105e01eb20c45

PR #704, opened 2025-08-18
Support @koa/router v14

timokoessler:


library/package.json
```diff
@@ -53,7 +53,7 @@
     "@hapi/hapi": "^21.3.10",
     "@hono/node-server": "^1.12.2",
     "@koa/bodyparser": "^6.0.0",
-    "@koa/router": "^13.0.0",
+    "@koa/router": "^14.0.0",
```
hansott, line 57: maybe we test multiple versions at this point? like the other frameworks?

timokoessler: Not really a big change https://github.com/koajs/router/releases/tag/v14.0.0, but can do

library/package.json
```diff
@@ -55,6 +55,11 @@
     "@hono/node-server": "^1.12.2",
     "@koa/bodyparser": "^6.0.0",
     "@koa/router": "^14.0.0",
+    "@koa/router-v10": "npm:@koa/router@^10.0.0",
+    "@koa/router-v11": "npm:@koa/router@^11.0.0",
+    "@koa/router-v12": "npm:@koa/router@^12.0.0",
+    "@koa/router-v13": "npm:@koa/router@^13.0.0",
+    "@koa/router-v14": "npm:@koa/router@^14.0.0",
```
timokoessler: As you keep @koa/router without version suffix, it is now duplicate?

hansott (approved):

PR #709, opened 2025-08-22
SQL early return: Allow dots in numbers

timokoessler:
+ Add test for metadata length


timokoessler:
Fixed in Zen internals instead

PR #715, opened 2025-08-28
perf: Check JWT min length

timokoessler:


library/helpers/tryDecodeAsJWT.test.ts
```diff
@@ -16,13 +16,13 @@ t.test("it returns payload for invalid JWT", async () => {
   // According to the JWT spec, the payload is not valid, but we'll extract the payload anyway
 
   // e30= is a base64 encoded string of '{}'
-  t.same(tryDecodeAsJWT("/;ping%20localhost;.e30=."), {
+  t.same(tryDecodeAsJWT("/;ping%20localhost---;.e30=."), {
```
timokoessler, line 19: The bypass here was that the JWT itself was not added to the context. This is covered by another test. The minimum length change does not allow this bypass again.

hansott (approved):

PR #734, opened 2025-09-09
Report stored SSRF as attack event

timokoessler:
image

### DIFF
--- a/library/vulnerabilities/ssrf/inspectDNSLookupCalls.ts
+++ b/library/vulnerabilities/ssrf/inspectDNSLookupCalls.ts
@@ -120,7 +120,7 @@ function wrapDNSLookupCallback(
       if (agent.shouldBlock()) {
         return callback(
           new Error(
-            `Zen has blocked ${attackKindHumanName("stored-ssrf")}: ${operation}(...) originating from unknown source`
+            `Zen has blocked ${attackKindHumanName("stored_ssrf")}: ${operation}(...) originating from unknown source`
           )
         );
       }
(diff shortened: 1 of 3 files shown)


library/vulnerabilities/ssrf/inspectDNSLookupCalls.ts
```diff
@@ -103,7 +103,26 @@ function wrapDNSLookupCallback(
...
+        paths: [],
+        metadata: getMetadataForSSRFAttack({
+          hostname,
+          port: undefined,
+          privateIP: imdsIpResult.ip,
+        }),
+        request: undefined,
+        payload: undefined,
+      });
+
```
willem-delbare, line 116: Why are we reporting before shouldblock?

timokoessler: We are reporting as detected then, like for every other attack. This should block is used for throwing the exception.

library/vulnerabilities/ssrf/inspectDNSLookupCalls.ts
```diff
@@ -174,6 +184,28 @@ function wrapDNSLookupCallback(
...
+          privateIp: imdsIpResult.ip,
+          callingLocationStackTrace,
+        });
+
+        // Block stored SSRF attack that target IMDS IP addresses
+        // An attacker could have stored a hostname in a database that points to an IMDS IP address
+        if (agent.shouldBlock()) {
+          return callback(
+            new Error(
+              `Zen has blocked ${attackKindHumanName("stored_ssrf")}: ${operation}(...) originating from unknown source`
```
hansott, line 203:
```suggestion
              `Zen has blocked ${attackKindHumanName("ssrf")}: ${operation}(...) originating from unknown source`
```

Isn't this a normal SSRF then?

timokoessler: No, stored, because of `if (!found) {`

PR #735, opened 2025-09-10
Use command.input.modelId instead of body.model

hansott:
Extract model ID from command input instead of response body to include the geography prefix

e.g. "us.anthropic.claude-3-5-sonnet-20240620-v1:0" instead of "claude-3-5-sonnet-20240620"

### DIFF
--- a/library/sinks/AwsSDKVersion3.ts
+++ b/library/sinks/AwsSDKVersion3.ts
@@ -77,12 +77,12 @@ export class AwsSDKVersion3 implements Wrapper {
         outputTokens = body.usage.output_tokens;
       }
 
+      let modelId: string | undefined = undefined;
       // @ts-expect-error We don't know the type of command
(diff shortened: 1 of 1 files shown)


library/sinks/AwsSDKVersion3.test.ts
```diff
@@ -45,22 +56,36 @@ t.test(
       }),
     });
 
-    const response = await client.send(command);
-    const responseBody = JSON.parse(new TextDecoder().decode(response.body));
+    await client.send(invokeCommand);
 
-    t.same(responseBody.content[0].text.trim(), "Paris");
-    t.match(agent.getAIStatistics().getStats(), [
-      {
```
timokoessler, line 53: Test assertions removed?

hansott: 4a464785725f1e1deb931fa31bd67bf1229665a7

PR #737, opened 2025-09-10
ESM: Add hooks self-check

timokoessler:


library/agent/hooks/instrumentation/checkHooks.register.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  async (t) => {
+    // This function already calls checkHooks, so we don't need to call it again
+    registerNodeHooks();
+
+    // Required because the check itself is async
+    await new Promise((resolve) => setTimeout(resolve, 100));
+
+    t.same(logs, []);
+  }
+);
```
hansott, line 27:
```suggestion
import * as t from "tap";
import { wrap } from "../../../helpers/wrap";
import { registerNodeHooks } from ".";
import * as mod from "node:module";
import { setTimeout } from "node:timers/promises";

const logs: string[] = [];
wrap(console, "warn", function warn() {
  return function warn(...args: string[]) {
    logs.push(...args);
  };
});

t.test(
  "it works",
  {
    skip: !("registerHooks" in mod) ? "Recent Node.js version required" : false,
  },
  async (t) => {
    // This function already calls checkHooks, so we don't need to call it again
    registerNodeHooks();

    // Required because the check itself is async
    await setTimeout(100);

    t.same(logs, []);
  }
);
```

hansott: but not so important

PR #743, opened 2025-09-16
Refactor DNS result inspection

timokoessler:
Remove double stored SSRF check through refactoring


timokoessler:
Reason for the missing lines in Code Coverage is that the redirection protection is still disabled.

PR #748, opened 2025-09-17
Extend libraries supported by the new instrumentation system

timokoessler:
- Undici
- GoogleGenAi
- GraphQL

- Support ArrowFunctions in Assignments
- Apply Array.from for args in case of modifyReturnValue

### DIFF
--- a/library/sinks/GoogleGenAi.ts
+++ b/library/sinks/GoogleGenAi.ts
@@ -125,7 +125,7 @@ export class GoogleGenAi implements Wrapper {
           {
             name: "this.generateContent",
             nodeType: "FunctionAssignment",
-            operationKind: undefined,
+            operationKind: "ai_op",
             modifyReturnValue: (args, returnValue, agent) => {
               this.inspectReturnValue(agent, returnValue);
               return returnValue;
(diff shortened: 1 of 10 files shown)


library/sinks/GoogleGenAi.ts
```diff
@@ -108,6 +118,20 @@ export class GoogleGenAi implements Wrapper {
...
         return exports;
-      });
+      })
+      .addMultiFileInstrumentation(
+        ["dist/node/index.cjs", "dist/node/index.mjs"],
+        [
+          {
+            name: "this.generateContent",
+            nodeType: "FunctionAssignment",
+            operationKind: undefined,
```
hansott:
```suggestion
            operationKind: "ai_op",
```

library/sinks/GoogleGenAi.ts
```diff
@@ -70,6 +70,16 @@ export class GoogleGenAi implements Wrapper {
             module: "@google/genai",
           });
         });
+    } else {
+      try {
+        this.inspectResponse(agent, returnValue);
```
hansott, line 77:
```suggestion
        // The new instrumentation wraps the return statements inside the async function
        // Inside an async function you can just return normal values (it's not a promise yet)
        this.inspectResponse(agent, returnValue);
```

PR #751, opened 2025-09-18
Add new instrumentation for bedrock client

hansott:


library/sinks/AwsSDKVersion3.ts
```diff
@@ -141,51 +142,63 @@ export class AwsSDKVersion3 implements Wrapper {
...
-              }
-
-              return returnValue;
-            },
-          });
-        }
+        this.onRequire(exports, pkgInfo);
+      })
+      .addFileInstrumentation({
+        path: "dist-cjs/index.js",
```
timokoessler, line 196: ESM missing? In that case we would need to modify the BedrockRuntimeClient class directly, not possible to access or modify exports. So maybe create a function that directly accepts the BedrockRuntimeClient?

hansott: See comment! https://github.com/AikidoSec/firewall-node/pull/751/commits/038e14d99c7bd7ea65327037989a7c96950efba2

PR #755, opened 2025-09-24
New instrumentation: Support AI SDK

timokoessler:
Also add support for return await.


library/sinks/AiSDK.ts
```diff
@@ -157,6 +158,16 @@ export class AiSDK implements Wrapper {
                 module: "ai",
               });
             });
+        } else {
+          try {
+            this.inspectAiCall(agent, args, returnValue);
+          } catch (error) {
+            agent.onErrorThrownByInterceptor({
+              error: error instanceof Error ? error : new Error(String(error)),
+              method: `${methodName}`,
```
hansott:
```suggestion
              method: methodName,
```

PR #758, opened 2025-09-25
Add Safe Chain

timokoessler:
+ Update GitHub actions
+ Run e2e tests on non EOL version
+ Update linter


.github/workflows/end-to-end-tests.yml
```diff
@@ -48,15 +48,19 @@ jobs:
     timeout-minutes: 15
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [22.x]
```
hansott: run on both?

PR #762, opened 2025-09-25
Add AWS SDK v2 instrumentation (and run missing tests)

hansott:


library/sinks/testLock.ts
```diff
@@ -0,0 +1,66 @@
+import { mkdir, open, unlink } from "fs/promises";
+import { setTimeout } from "timers/promises";
+import * as fs from "fs/promises";
```
timokoessler: Maybe add a single type import for FileHandle?

timokoessler (approved):

library/sinks/testLock.ts
hansott: https://github.com/AikidoSec/firewall-node/pull/762/commits/640a4999a020c59021dcec0a2fc72fd1b2e1faa2

PR #765, opened 2025-09-28
Add support for multi environment tokens (EU/US/ME)

hansott:
Closes AIK-6819


library/agent/extractRegionFromToken.ts
```diff
@@ -0,0 +1,16 @@
+export function extractRegionFromToken(token: string): string {
+  if (!token || !token.startsWith("AIK_RUNTIME_")) {
```
timokoessler, line 2: To confirm, the tokens are case sensitive?

PR #771, opened 2025-10-03
Send startup event during first invocation (serverless)

hansott:


library/sources/FunctionsFramework.test.ts
```diff
@@ -208,3 +210,79 @@ t.test("it hooks into functions framework", async () => {
 
   t.same(isWrapped(framework.http), true);
 });
+
+t.test("it waits for attack events to be sent before returning", async (t) => {
+  let attackReportResolveCount = 0;
+
+  const api = new ReportingAPIForTesting();
+
+  wrap(api, "report", function report(original) {
```
timokoessler, line 270: Not possible to use api.getEvents()?

hansott: Thanks! https://github.com/AikidoSec/firewall-node/pull/771/commits/2f7e0a855f7073c0521c6de673fa817691ee3a04

library/sources/FunctionsFramework.ts
```diff
@@ -1,45 +1,108 @@
...
         routeParams: {},
         source: "cloud-function/http",
-        route: undefined,
+        route: buildRouteFromURL(url),
       },
       async () => {
+        let result: unknown;
         try {
-          return await fn(req, res);
+          result = await fn(req, res);
```
timokoessler: result is not used anywhere else?

library/sources/FunctionsFramework.ts
```diff
@@ -1,45 +1,108 @@
...
+    if (!isNaN(parsed) && parsed >= 60 * 1000) {
+      return parsed;
+    }
+  }
+
+  return 10 * 60 * 1000; // 10 minutes
+}
+
+function getTimeoutInMS(): number {
+  if (process.env.AIKIDO_CLOUD_FUNCTION_TIMEOUT_MS) {
```
timokoessler, line 27: Might be nice to have a test that actually sets the env vars?

hansott: https://github.com/AikidoSec/firewall-node/pull/771/commits/61739758dd7de21a7d71f2c4fb711a6a4cecbec9
Good idea, added some tests to cover logic inside these functions (at least)

library/sources/FunctionsFramework.ts
hansott: https://github.com/AikidoSec/firewall-node/pull/771/commits/c17297a8f7ad4ebb55517bffebab82bc93ee2e90
Oops, good catch

PR #778, opened 2025-10-16
Node.js v25

timokoessler:
+ Fix express benchmark calculations


.github/workflows/lint-code.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [18.x]
+        node-version: [24.x]
```
timokoessler, line 9: Because 18 is EOL and 24 is now the active release line.

hansott (approved):

PR #792, opened 2025-10-30
Add Sentry ESM e2e test

timokoessler:


scripts/install.js
```diff
@@ -45,6 +46,13 @@ async function main() {
 async function installDependencies(folder) {
   console.log(`Installing dependencies for ${folder}`);
 
+  if (folder === "sample-apps/n8n" && majorNodeVersion >= 24) {
```
hansott: Perhaps create an array above? e.g.

```js
const skipInstall = [{ folder: "sample-apps/n8n", check: (major) => major >= 24 }]
```

PR #802, opened 2025-10-31
perf: Avoid for loop over sources

hansott:
It's a tiny improvement but makes it a bit more readable too.

### DIFF
--- a/library/agent/Context.ts
+++ b/library/agent/Context.ts
@@ -56,8 +56,9 @@ export function updateContext<K extends keyof Context>(
 ) {
   context[key] = value;
 
-  // Clear all the cached user input strings
   if (isSourceKey(key)) {
+    // Clear all the cached user input strings
+    // Only if user input related fields are updated
     delete context.cache;
   }
 }


library/vulnerabilities/ssrf/findHostnameInContext.ts
```diff
@@ -39,15 +40,11 @@ export function findHostnameInContext(
...
-  for (const source of SOURCES) {
-    const userInput = extractStringsFromUserInputCached(context, source);
-    if (!userInput) {
-      continue;
-    }
-
-    for (const str of userInput) {
-      const found = findHostnameInUserInput(str, hostname, port);
-      if (found) {
+  for (const str of extractStringsFromUserInputCached(context)) {
```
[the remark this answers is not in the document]
hansott: Always returns `Set<string>`

library/agent/Context.ts
```diff
@@ -56,9 +51,8 @@ export function updateContext<K extends keyof Context>(
 ) {
   context[key] = value;
 
-  if (context.cache && isSourceKey(key)) {
-    context.cache.delete(key);
-  }
+  // Clear all the cached user input strings
```
timokoessler: Do you know how big the performance impact of this change is? I think we are calling updateContext at least twice per request. But understand why you changed it. If we would still store the sources separate in the cache, we would still need a loop tho.

hansott: https://github.com/AikidoSec/firewall-node/pull/802/commits/ec235add1106233dc5dc11a01c2895bb564ea41f

PR #803, opened 2025-10-31
Use separate cache for path traversal check

hansott:
We have to check every user input strings. We can reduce the amount of strings with a separate cache specifically for file system checks. The string needs to include at least a slash. Otherwise it can be never be abused.

Something we could apply to other vulnerability checks :)

### DIFF
--- a/library/helpers/extractPathStringsFromUserInputCached.ts
+++ b/library/helpers/extractPathStringsFromUserInputCached.ts
@@ -20,7 +20,7 @@ export function extractPathStringsFromUserInputCached(
     }
 
     for (const item of extractStringsFromUserInput(context[source])) {
-      // Performance optimization: only keep strings that contain a slash
+      // Performance optimization: only keep strings that contain a path separator
       // as only those can be used for path traversal
       // keeps the set smaller and speeds up `fs` and `path` operations
       if (item.includes(sep)) {


timokoessler (approved):

library/helpers/extractPathStringsFromUserInputCached.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (item.includes("/")) {
+        userStrings.add(item);
+      }
+    }
+  }
+
+  context.cachePathTraversal = userStrings;
+
+  return userStrings;
+}
```
hansott, line 35:
```suggestion
import { extractStringsFromUserInput } from "./extractStringsFromUserInput";
import path from "node:path";

type ReturnValue = ReturnType<typeof extractStringsFromUserInput>;

export function extractPathTraversalStringsFromUserInputCached(
  context: Context
): ReturnValue {
  if (context.cachePathTraversal) {
    return context.cachePathTraversal;
  }

  const userStrings: ReturnValue = new Set();

  for (const source of SOURCES) {
    if (!context[source]) {
      continue;
    }

    for (const item of extractStringsFromUserInput(context[source])) {
      // Performance optimization: only keep strings that contain a path separator
      // as only those can be used for path traversal
      // keeps the set smaller and speeds up `fs` and `path` operations
      if (item.includes(path.sep)) {
        userStrings.add(item);
      }
    }
  }

  context.cachePathTraversal = userStrings;

  return userStrings;
}
```

hansott: https://github.com/AikidoSec/firewall-node/pull/803/commits/425f460b17658de399d5e4553a870945a20708d5

library/helpers/extractPathStringsFromUserInputCached.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  const userStrings: ReturnValue = new Set();
+
+  for (const source of SOURCES) {
+    if (!context[source]) {
+      continue;
+    }
+
+    for (const item of extractStringsFromUserInput(context[source])) {
+      // Performance optimization: only keep strings that contain a slash
```
hansott:
```suggestion
      // Performance optimization: only keep strings that contain a path separator
```

PR #806, opened 2025-11-05
troubleshooting draft

kidk:


docs/troubleshooting.md
```diff
@@ -0,0 +1,80 @@
+# Troubleshooting
+
+## Check logs for errors
+
+Common places:
+- Docker: `docker logs <your-app-container>`
+- systemd: `journalctl -u <your-app-service> --since "1 hour ago"`
```
timokoessler, line 23: Not really common to run Node.js apps as systemd service. But still possible tho.

docs/troubleshooting.md
```diff
@@ -0,0 +1,80 @@
...
+- Local dev: your terminal or IDE run console
+
+Tip: search for lines that contain `Aikido` or `Zen` to spot initialization and request logs.
+
+## Confirm the dependency is present
```
timokoessler: Maybe enough to confirm if it is imported? Otherwise they will get an exception saying that the package was not found.

docs/troubleshooting.md
```diff
@@ -0,0 +1,80 @@
...
+```
+pnpm ls | grep -i aikido
+cat package.json | grep -i aikido
+```
+
+## Confirm middleware is registered early
+
+Register the firewall as high as possible in the request pipeline so it sees every request.
+
+**Express**
```
timokoessler: Link to each framework doc instead? Right now only some frameworks are mentioned and we would need to update in two places.

docs/troubleshooting.md
```diff
@@ -0,0 +1,80 @@
...
+```
+// main.ts, before app.listen
+import { NestFactory } from "@nestjs/core";
+import aikido from ".../aikido";
+
+const app = await NestFactory.create(AppModule);
+app.use(aikido());
+await app.listen(3000);
+```
+
```
timokoessler, line 35:
```suggestion
## Enable debug logging

Set the environment variable `AIKIDO_DEBUG` to `true` and check the log output of your application.
You should see the message `AIKIDO: Starting agent vX.X.X`.
In addition the output contains the name and version of each supported and instrumented library or framework in the format `library@version is supported!`. Please ensure that the logs contain these message at least for your web framework (e.g. express) and your database driver.

```

timokoessler, line 35: I think we should also mention ESM at runtime and bundlers?

docs/troubleshooting.md
```diff
@@ -0,0 +1,43 @@
...
+You should see the message `AIKIDO: Starting agent vX.X.X`.
+
+In addition the output contains the name and version of each supported and instrumented library or framework in the format `library@version is supported!`. Please ensure that the logs contain these message at least for your web framework (e.g. express) and your database driver.
+
+## Contact support
+
+If you still can’t resolve the issue:
+
+- Use the in-app chat to reach our support team directly.
```
hansott:
```suggestion
- Use the [in-app chat](https://app.aikido.dev/) to reach our support team directly.
```

docs/troubleshooting.md
```diff
@@ -0,0 +1,43 @@
+# Troubleshooting
+
+## Review installation steps
+
+Double-check your setup against the [installation guide](../README.md#installation).  
+Make sure:
+- The package installed correctly.  
+- The firewall is initialized early in your app (before routes or handlers).
```
hansott: should be before any import, any import before might not be protected!

docs/troubleshooting.md
```diff
@@ -0,0 +1,43 @@
+# Troubleshooting
+
+## Review installation steps
+
+Double-check your setup against the [installation guide](../README.md#installation).  
+Make sure:
+- The package installed correctly.  
+- The firewall is initialized early in your app (before routes or handlers).  
+- Your framework-specific integration (middleware, decorator, etc.) matches the example in the README.  
+- You’re running a supported runtime version for your language.
```
hansott:
```suggestion
- You’re using Node.js 16 or newer
```

docs/troubleshooting.md
hansott: agree

docs/troubleshooting.md
hansott: yess please

docs/troubleshooting.md
timokoessler:
```suggestion
- The firewall is imported early in your app (before any other import).
```

docs/troubleshooting.md
```diff
@@ -0,0 +1,43 @@
+# Troubleshooting
+
+## Review installation steps
+
+Double-check your setup against the [installation guide](../README.md#installation).  
+Make sure:
+- The package installed correctly.  
+- The firewall is initialized early in your app (before routes or handlers).  
+- Your framework-specific integration (middleware, decorator, etc.) matches the example in the README.  
+- You’re using Node.js 16 or newer
```
timokoessler:
```suggestion
- You’re using Node.js 16 or newer
- Zen runs only on the server side, it does not run in the browser.
```

docs/troubleshooting.md
timokoessler:
```suggestion

## Bundlers & ECMAScript modules (ESM)

If you are using a bundler like esbuild or a framework that uses bundling please ensure to follow the steps described in the [bundler](./bundler.md) docs. Applications using ESM at runtime are not supported yet. TypeScript code is often translated to CommonJS.
```

docs/troubleshooting.md
```diff
@@ -0,0 +1,48 @@
+# Troubleshooting
+
+## Review installation steps
+
+Double-check your setup against the [installation guide](../README.md#installation).  
+Make sure:
+- The package installed correctly.  
+- The firewall is imported early in your app (before any other import).
+- Your framework-specific integration (middleware, decorator, etc.) matches the example in the README.  
+- You’re using Node.js 16 or newer
```
hansott:
```suggestion
- You’re using Node.js 16 or newer.
```

PR #816, opened 2025-11-12
Improve path traversal detection (URL)

hansott:


sample-apps/express-mysql/app.js
```diff
@@ -126,6 +128,17 @@ async function main(port) {
     }
   });
 
+  app.get("/search", (req, res) => {
+    const searchTerm = req.query.q;
+    const fileUrl = new URL(`file:///public/${searchTerm}`);
```
bitterpanda63: shouldnt we test with file://public/../../../ ?

PR #820, opened 2025-11-13
Skip build in unit tests if not necessary

timokoessler:
+ Fix npm ci on v25

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -70,6 +70,7 @@ jobs:
       - name: Start containers
         run: npm run containers
       - name: Prepare WASM components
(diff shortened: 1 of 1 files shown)


.github/workflows/unit-test.yml
```diff
@@ -65,14 +69,19 @@ jobs:
       - run: npm run install-lib-only
       - name: Start containers
         run: npm run containers
-      - run: npm run build
+      - name: Prepare WASM components
+        run: npm run build -- --only-wasm
```
hansott, line 74:
```suggestion
        # When running tests with tap (CJS), we don't need to build our lib, just need the WASM files
        run: npm run build -- --only-wasm
```

PR #821, opened 2025-11-13
Allow RegExp pattern for targeting bundled files (e.g. chunk-XXXX.mjs)

hansott:


library/agent/hooks/instrumentation/instructions.ts
```diff
@@ -117,6 +117,25 @@ export function shouldPatchPackage(name: string): boolean {
...
+  filePath: string
+): boolean {
+  if (typeof pathOrPattern === "string") {
+    return pathOrPattern === filePath;
+  }
+
+  return pathOrPattern.test(filePath);
+}
+
+function getIdentifier(pathOrPattern: string | RegExp) {
```
timokoessler, line 131: Does JS not call .toString() automatically?
```js
const regex: RegExp = /abcd/i;
console.log(`test${regex}A42`);
// Prints 'test/abcd/iA42'

```

PR #826, opened 2025-11-17
Downgrade npm for package lock bug on node v24

hansott:


.github/workflows/end-to-end-tests.yml
```diff
@@ -68,13 +68,9 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: "npm"
           cache-dependency-path: "**/package-lock.json"
-      - name: Setup Aikido safe-chain
```
timokoessler, line 71: Why did you remove Safe Chain here?

timokoessler (approved):

PR #827, opened 2025-11-18
Collect samples for attack wave detection

hansott:


library/vulnerabilities/attack-wave-detection/AttackWaveDetector.ts
```diff
@@ -2,8 +2,13 @@ import { LRUMap } from "../../ratelimiting/LRUMap";
...
 import { isWebScanner } from "./isWebScanner";
 
+export type SuspiciousRequest = {
+  method: string;
+  url: string;
+};
+
 export class AttackWaveDetector {
-  private suspiciousRequestsMap: LRUMap<string, number>;
+  private suspiciousRequestsMap: LRUMap<string, SuspiciousRequest[]>;
```
timokoessler: Only an idea: If we would use a separate Set, we would not store or show duplicates and we can keep the amount of samples independent from the count needed for triggering the event, if needed in the future.

hansott: Good idea!

timokoessler: We can't use a Set because they don't work on objects :(

library/vulnerabilities/attack-wave-detection/AttackWaveDetector.ts
```diff
@@ -72,4 +96,28 @@ export class AttackWaveDetector {
...
+    if (
+      samples.some(
+        (sample) =>
+          sample.method === request.method && sample.url === request.url
+      )
+    ) {
+      return;
+    }
+    samples.push(request);
+    this.suspiciousRequestsSamples.set(ip, samples);
```
hansott, line 121: I think we don't need the `.set` call because we modify the returned array? But let's keep it for making it obvious?

PR #837, opened 2025-11-26
POC for custom route patterns

hansott:


library/helpers/buildRouteFromURL.ts
```diff
@@ -86,5 +114,11 @@ function replaceURLSegmentWithParam(segment: string) {
     return ":secret";
   }
 
+  for (const pattern of customPatterns) {
+    if (pattern.test(segment)) {
+      return `:custom`;
```
tudor-timcu, line 119: shouldn't we return the actual pattern name set by the app? eg: :tenant

library/agent/addRouteParam.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const regex = compileCustomPattern(pattern);
+  if (!regex) {
+    console.warn(
+      "addRouteParam(...) could not compile the provided pattern into a valid regular expression."
+    );
+    return;
+  }
+
+  if (registeredPatternsSet.has(pattern)) {
```
timokoessler, line 28: Check before compiling?

library/helpers/buildRouteFromURL.ts
```diff
@@ -15,7 +17,7 @@ const HASH = /^(?:[a-f0-9]{32}|[a-f0-9]{40}|[a-f0-9]{64}|[a-f0-9]{128})$/i;
 const HASH_LENGTHS = [32, 40, 64, 128];
 const NUMBER_ARRAY = /^\d+(?:,\d+)*$/;
 
-export function buildRouteFromURL(url: string) {
+export function buildRouteFromURL(url: string, custom: RegExp[]) {
```
timokoessler, line 20: Make custom parameter optional and call getRegisteredRouteParams() by default, so we can still easily test but don't have to modify so many files? Only an idea, not required.

PR #839, opened 2025-11-27
Use built-in RegExp.escape if available

timokoessler:


library/helpers/escapeStringRegexp.ts
```diff
@@ -6,5 +6,13 @@
  * Taken from https://github.com/sindresorhus/escape-string-regexp/
  */
 export function escapeStringRegexp(string: string) {
+  // Use native implementation if available (Node.js 24+)
+  // A benchmark showed that it is up to 50% faster than the polyfill
+  // @ts-expect-error Outdated Node.js types
+  if (typeof RegExp?.escape === "function") {
+    // @ts-expect-error Outdated Node.js types
+    return RegExp.escape(string);
+  }
```
hansott: I don't think the `typeof` check will have a lot of overhead but we do this for `tryParseURL`? I do like this approach more because it's a bit more readable than the `canParse` stuff

```ts
function canParse(url: string): URL | undefined {
  if (URL.canParse(url)) {
    return new URL(url);
  }

  return undefined;
}

// URL.canParse(...) is a lot faster than using the constructor and catching the error
// Added in Node.js: v19.9.0, v18.17.0
if (typeof URL.canParse === "function") {
  tryParseURL = canParse;
}

export { tryParseURL };
```

timokoessler: Done, totally missed this :D

hansott (approved):

PR #841, opened 2025-11-28
Fix duplicate prettier install & format

timokoessler:


.github/workflows/lint-code.yml
```diff
@@ -33,7 +33,10 @@ jobs:
         run: npm i -g npm@11.6.0
       - run: npm run install-lib-only
       - run: npm run build
-      - run: npm run lint
+      - name: Run Linter for JavaScript/TypeScript
+        run: npm run lint
+      - name: Check formatting
+        run: npm run check:format
```
hansott:
```suggestion
        run: npm run format:check
```

So that the commands can be the same: `npm run format` / `npm run format:check`?

package.json
```diff
@@ -25,6 +25,7 @@
     "end2end:new": "cd end2end && npm run test:new",
     "format": "prettier --write .",
     "benchmark": "node scripts/benchmark.js",
-    "sample-app": "node scripts/run-sample-app.js"
+    "sample-app": "node scripts/run-sample-app.js",
+    "check:format": "prettier --check ."
```
hansott:
```suggestion
    "format:check": "prettier --check ."
```

PR #842, opened 2025-12-01
Add hook to monitor outbound requests

hansott:


library/agent/hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+    hooks.get(name)!.add(fn);
+  }
+}
+
+export function removeHook<N extends HookName>(
+  name: N,
+  fn: (...args: HookTypes[N]["args"]) => void | Promise<void>
+) {
+  if (hooks.has(name)) {
+    hooks.get(name)!.delete(fn);
```
timokoessler: Maybe this is simpler? Just an idea.
```ts
hooks.get(name)?.delete(fn);
```

timokoessler (approved):

library/agent/hooks.ts
hansott: https://github.com/AikidoSec/firewall-node/pull/842/commits/91c09e3c482a4f38e8c2b5110bc060a2281028c8

PR #843, opened 2025-12-04
Fix missing outbound hostname

timokoessler:
For internal http calls in ESM mode

### DIFF
--- a/end2end/tests-new/hono-pg-ts-esm.test.mjs
+++ b/end2end/tests-new/hono-pg-ts-esm.test.mjs
@@ -11,9 +11,6 @@ const pathToAppDir = resolve(
 );
 const port = await getRandomPort();
 const port2 = await getRandomPort();
-const port3 = await getRandomPort();
-
-const testServerUrl = "http://localhost:5874";
 
 test("it blocks request in blocking mode", async () => {
   const server = spawn(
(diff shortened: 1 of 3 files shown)


end2end/tests-new/hono-pg-ts-esm.test.mjs
```diff
@@ -147,3 +150,86 @@ test("it does not block request in monitoring mode", async () => {
...
+
+    const heartbeatEvent = heartbeatEvents[0];
+
+    equal(heartbeatEvent.hostnames.length, 1);
+
+    const hostname = heartbeatEvent.hostnames[0];
+
+    equal(hostname.hostname, "localhost");
+    equal(hostname.hits, 2);
+    equal(hostname.port, 5874);
```
hansott: and the real time polling happens after 1 minute, extend waiting time? So 3 hits?

(but it will be the same hostname + port though)

end2end/tests-new/hono-pg-ts-esm.test.mjs
```diff
@@ -147,3 +150,86 @@ test("it does not block request in monitoring mode", async () => {
...
+    // Wait for the server to start
+    await timeout(2000);
+
+    await fetch(`http://127.0.0.1:${port3}/`, {
+      method: "GET",
+      signal: AbortSignal.timeout(5000),
+    });
+
+    // Wait for heartbeat to be sent
+    await timeout(35000);
```
hansott:
```suggestion
    await timeout(60000);
```

PR #844, opened 2025-12-05
Convert e2e test mock server to Typescript

timokoessler:
I noticed multiple tiny issues in the e2e test server, e.g. app.serviceId was used instead of app.id. This PR fixes this and also converts it to TypeScript so we notice things like this in the future.

### DIFF
--- a/end2end/server/src/zen/apps.ts
+++ b/end2end/server/src/zen/apps.ts
@@ -4,7 +4,7 @@ export type App = {
   id: number;
   token: string;
   configUpdatedAt: number;
-}
+};
 
 const apps: App[] = [];
 
--- a/end2end/server/.dockerignore
+++ b/end2end/server/.dockerignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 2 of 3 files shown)


end2end/server/src/zen/apps.ts
```diff
@@ -1,47 +1,42 @@
-const { randomInt, timingSafeEqual } = require("crypto");
+import { randomInt, timingSafeEqual } from "node:crypto";
 
-const apps = [];
+export interface App {
```
hansott:
```suggestion
export type App = {
```

end2end/server/package.json
```diff
@@ -1,9 +1,19 @@
...
     "express": "^5.1.0"
+  },
+  "devDependencies": {
+    "@types/express": "^5.0.6",
+    "@types/node": "^24.10.1",
+    "typescript": "^5.9.3"
+  },
+  "scripts": {
+    "start": "node ./app.ts",
+    "check:types": "tsc --noEmit"
```
hansott, line 17: we don't run this in CI, right? should we add it to the Dockerfile? When it's building?

PR #851, opened 2025-12-12
Only build once during CI

timokoessler:
Also includes multiple speed ups.


library/helpers/isLibBundled.ts
```diff
@@ -5,6 +5,6 @@ export function isLibBundled(): boolean {
 
   return (
     !normalizedDirName.includes("node_modules/@aikidosec/firewall/helpers") &&
-    !normalizedDirName.includes("firewall-node/build/helpers") // In case of e2e tests
+    !normalizedDirName.includes("/build/helpers") // In case of e2e tests
```
hansott, line 8: related change?

timokoessler: Yes, because GitHub changed the name how he checkouts the repo, I guess because it runs more things in parallel. Will be further changed again in #812 tho

PR #1049, opened 2026-06-01
Support ip based rate limiting for Next.js

timokoessler:
## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Implemented IP-based rate limiting for Next.js, returned 429 with Retry-After

**🔧 Refactors**
* Renamed readBody flag to readBodyAndRateLimit and threaded flag

<sup>[More info](https://app.aikido.dev/featurebranch/scan/130535786?groupId=6)</sup>

### DIFF
--- a/library/sources/HTTPServer.test.ts
+++ b/library/sources/HTTPServer.test.ts
@@ -1231,6 +1231,63 @@ t.test("It ignores multipart form data files", async (t) => {
   });
 });
 
+t.test("rate limits requests when NEXT_DEPLOYMENT_ID is set", async (t) => {
+  process.env.NEXT_DEPLOYMENT_ID = "";
+
+  const server = http.createServer((req, res) => {
+    res.statusCode = 200;
+    res.end("OK");
+  });
+
+  await new Promise<void>((resolve) => {
+    server.listen(3233, async () => {
+      const req = () =>
+        fetch({
+          url: new URL("http://localhost:3233/rate-limited"),
+          method: "GET",
+          headers: { "x-forwarded-for": "100.0.0.1" },
+          timeoutInMS: 500,
+        });
+
+      t.equal((await req()).statusCode, 200);
+      t.equal((await req()).statusCode, 200);
+
+      const blocked = await req();
+      t.equal(blocked.statusCode, 429);
+      t.match(blocked.body, /You are rate limited by Zen/);
+
+      server.close();
+      resolve();
+    });
+  });
+});
+
+t.test("does not rate limit when NEXT_DEPLOYMENT_ID is not set", async (t) => {
+  const server = http.createServer((req, res) => {
+    res.statusCode = 200;
+    res.end("OK");
+  });
+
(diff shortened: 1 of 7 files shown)


library/sources/HTTPServer.test.ts
```diff
@@ -1231,6 +1231,63 @@ t.test("It ignores multipart form data files", async (t) => {
   });
 });
 
+t.test("rate limits requests when NEXT_DEPLOYMENT_ID is set", async (t) => {
+  process.env.NEXT_DEPLOYMENT_ID = "";
```
tomaisthorpe: I think this is fine, but is it not better giving it an actual value here? I think sometimes an empty value could be confused for it not being set?

library/sources/http-server/createRequestListener.ts
```diff
@@ -84,10 +97,33 @@ function callListenerWithContext(
       return;
     }
 
+    if (shouldRateLimit && applyRateLimiting(res)) {
```
hansott, line 102:
```suggestion
    // Rate limiting normally happens in our framework specific middleware (e.g. express)
    // For certain frameworks, e.g. next.js, we need to do this here so that it works out of the box
    if (shouldRateLimit && applyRateLimiting(res)) {
```

library/sources/http-server/createRequestListener.ts
```diff
@@ -84,10 +97,33 @@ function callListenerWithContext(
...
+    }
+
     return listener(req, res);
   });
 }
 
+function applyRateLimiting(res: ServerResponse): boolean {
+  const result = shouldBlockRequest();
+
+  if (result.block && result.type === "ratelimited") {
```
hansott, line 114:
```suggestion
  // We don't support blocked users here because the user's code hasn't been called yet
  if (result.block && result.type === "ratelimited") {
```

library/sources/http-server/createRequestListener.ts
```diff
@@ -25,10 +27,19 @@ export function createRequestListener(
     // This is tricky, see replaceRequestBody(...)
     // e.g. Hono uses web requests and web streams
     // (uses Readable.toWeb(req) to convert to a web stream)
-    const readBody = "NEXT_DEPLOYMENT_ID" in process.env || isMicroInstalled;
-
-    if (!readBody) {
-      return callListenerWithContext(listener, req, res, module, agent, "");
+    const readBodyAndRateLimit =
+      "NEXT_DEPLOYMENT_ID" in process.env || isMicroInstalled;
```
hansott, line 31: this will also be used for micro but we don't have any tests for micro atm?

timokoessler: Added e2e test in 801fb45eca82847535c96515c08be6cd927b7084

hansott (approved):

PR #1105, opened 2026-07-30
Fix IDOR protection edge case for MySQL2

timokoessler:


library/sinks/MySQL2.ts
```diff
@@ -67,9 +70,23 @@ export class MySQL2 implements Wrapper {
...
+          : undefined;
+      const argValues = this.findParams(args);
+
+      // mysql2's query() lets the second argument's values override the object's
+      // values, while execute() (and prepare()) prefer the object's values over
+      // the second argument's values.
+      const params =
+        operation === "mysql2.query"
+          ? (argValues ?? objectValues)
+          : (objectValues ?? argValues);
```
hansott: maybe extract method or use proper if? the ternary makes it a bit more difficult to read?

timokoessler: d3a08527fbef97dc6bfa3ce1ea4d5a50dd73c8c2

hansott (approved):

PR #1109, opened 2026-07-31
Fix AWS Gateway event v2 with removed cookie header

timokoessler:


timokoessler:
> 0 Missing and 1 partial ⚠️

Not possible to trigger this partial missing test coverage because events without headers are already ignored before, but still an additional safety check.

hansott (approved):

PR #1113, opened 2026-08-05
Print warning when using @react-router/serve

hansott:


end2end/tests-new/react-router-serve-warning.test.mjs
```diff
@@ -0,0 +1,38 @@
...
+
+  let stderr = "";
+  server.stderr.on("data", (data) => {
+    stderr += data.toString();
+  });
+
+  await timeout(3000);
+
+  match(stderr, /@react-router\/serve/);
+  match(stderr, /not supported yet/);
```
timokoessler, line 35: Might be nice to also add a not matching test to some existing e2e test?

timokoessler (approved):

end2end/tests-new/react-router-serve-warning.test.mjs
hansott: 834f5f80cd970366ff4da9690ffee86c0c41bdcd

timokoessler (approved):

PR #1126, opened 2026-08-14
Reduce blocking of event loop when updating ip lists

timokoessler:
Simpler approach than #1122


timokoessler:
Replaced by https://github.com/AikidoSec/zen-internals-node/pull/11

